PR #5, opened 2016-09-05
MO: Bail out when PS version is lower than 1.7

firstred:


maximebiloe:
In my opinion, it won't prevent the blank page cause by a parsing error (not execution) with the `use` statement on PHP < 5.3. It will just prevent 1.6 PrestaShop running on PHP >= 5.3 from installing the module which is already prevented with the `ps_versions_compliancy`.
So I close this PR waiting for a better solution.
Thanks

PR #7, opened 2016-09-13
Rename JS file & fix layout

maximebiloe:


vincentbz:
Tested and ok, thanks !

PR #20, opened 2017-06-29
Display products only with associated category

fatmaBouchekoua:
This PR should fix the problem of displaying products when selecting parent categories in the FO even when they are not associated to these categories.

Fixed ticket : http://forge.prestashop.com/browse/BOOM-3111

### DIFF
--- a/ps_facetedsearch.php
+++ b/ps_facetedsearch.php
@@ -1782,11 +1782,6 @@ class Ps_Facetedsearch extends Module implements WidgetInterface
                             ));
                         }
 
-                        $depth = Configuration::get('PS_LAYERED_FILTER_CATEGORY_DEPTH');
-                        if ($depth === false) {
-                            $depth = 1;
-                        }
-
                         $sql_query['select'] = '
                         SELECT c.id_category, c.id_parent, cl.name, (SELECT count(DISTINCT p.id_product) # ';
                         $sql_query['from']   = '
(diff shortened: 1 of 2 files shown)


prasanthSelvar:
Tested and OK, Thanks Fatma

aleeks:
Like it was before, I understand the param `PS_LAYERED_FILTER_CATEGORY_DEPTH` was used to retrieve category with level depth, so for example: 

home 
    women
        tops
            t-shit
            tops
        dresses
            casual
            evening
            summer

if PS_LAYERED_FILTER_CATEGORY_DEPTH = 2, summer have to be retrieved if it's dresses & women.

With the solution proposed, summer is only retrieve with dresses because it's it parents, but not with women.

I don't know if i'm clear but.. maybe it was thinking like that.

ping @maximebiloe @Quetzacoalt91 @eternoendless @jocel1

eternoendless:
Ping @jocel1

jocel1:
I also confirm it looks really suspicious. To fix it properly it seems we need to check if there's any "associated categories" for a given parent->id, and only if this is the case, use the list of those associated categories for c.id_category IN (id list)

vincentbz:
hi,

Well, this is actually not a bug but a question of setting. In the faceted navigation module, you can select if you want to display the products of the subcategories or not. 
I close this PR, thanks guys !

PR #24, opened 2017-09-26
Product sorting 

manfield:
Fixed product sorting (according to position) when processed by ps_facetedsearch module as Product Search provider.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/ps_facetedsearch/24)

### DIFF
--- a/ps_facetedsearch.php
+++ b/ps_facetedsearch.php
@@ -1438,9 +1438,9 @@ class Ps_Facetedsearch extends Module implements WidgetInterface
         Db::getInstance()->execute('DROP TEMPORARY TABLE IF EXISTS '._DB_PREFIX_.'cat_filter_restriction', false);
         if (empty($selected_filters['category'])) {
             /* Create the table which contains all the id_product in a cat or a tree */
-            Db::getInstance()->execute('CREATE TEMPORARY TABLE '._DB_PREFIX_.'cat_filter_restriction ENGINE=MEMORY
-                                                        SELECT cp.id_product, MIN(cp.position) position FROM '._DB_PREFIX_.'category c
-                                                        STRAIGHT_JOIN '._DB_PREFIX_.'category_product cp ON (c.id_category = cp.id_category AND
+            Db::getInstance()->execute('CREATE TEMPORARY TABLE '._DB_PREFIX_.'cat_filter_restriction ENGINE=MEMORY 
+                                                        SELECT cp.id_product, cp.position FROM '._DB_PREFIX_.'category_product cp 
+                                                        STRAIGHT_JOIN '._DB_PREFIX_.'category c ON (c.id_category = cp.id_category AND
                                                         '.($this->ps_layered_full_tree ? 'c.nleft >= '.(int) $parent->nleft.'
                                                         AND c.nright <= '.(int) $parent->nright : 'c.id_category = '.(int) $id_parent).'
                                                         AND c.active = 1)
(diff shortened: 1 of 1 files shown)


tomlev:
Reviewed 1 of 1 files at r1.
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---

*[ps_facetedsearch.php, line 1548 at r1](https://reviewable.io:443/reviews/prestashop/ps_facetedsearch/24#-KwARDmrqfR0dE5r-e-e:-KwARDmrqfR0dE5r-e-f:b-52u44q) ([raw file](https://github.com/prestashop/ps_facetedsearch/blob/03aa3b85f31d434ff3d325e406d05742a2b16dfb/ps_facetedsearch.php#L1548)):*
> ```PHP
>                     ' LIMIT '.(((int) $page - 1) * $products_per_page.','.$products_per_page), true, false);
>                     
>                     
> ```

useless spaces (not blocker)

---

*[ps_facetedsearch.php, line 1589 at r1](https://reviewable.io:443/reviews/prestashop/ps_facetedsearch/24#-KwARHlG5ABWJZVjR8i_:-KwARHlG5ABWJZVjR8ia:b-52u44q) ([raw file](https://github.com/prestashop/ps_facetedsearch/blob/03aa3b85f31d434ff3d325e406d05742a2b16dfb/ps_facetedsearch.php#L1589)):*
> ```PHP
>             'count' => $this->nbr_products,
>         );
>       
> ```

useless spaces (not blocker)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/prestashop/ps_facetedsearch/24)*

tomlev:
thank you @manfield

---

Review status: all files reviewed at latest revision, 2 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/prestashop/ps_facetedsearch/24#-:-KwARwVswiBK0sxOnoiZ:b-8ku5bw)*

jocel1:
I'll check if I can integrate the changes directly into https://github.com/PrestaShop/ps_facetedsearch/pull/25 since it changes the same lines.

ps_facetedsearch.php
```diff
@@ -1438,27 +1438,27 @@ public function getProductByFilters(
...
+                                                        STRAIGHT_JOIN '._DB_PREFIX_.'category c ON (c.id_category = cp.id_category AND
                                                         '.($this->ps_layered_full_tree ? 'c.nleft >= '.(int) $parent->nleft.'
                                                         AND c.nright <= '.(int) $parent->nright : 'c.id_category = '.(int) $id_parent).'
                                                         AND c.active = 1)
                                                         STRAIGHT_JOIN `'._DB_PREFIX_.'product` p ON (p.id_product=cp.id_product)
                                                         '.$price_filter_query_in.'
                                                         '.$query_filters_from.'
                                                         WHERE 1 '.$query_filters_where.'
-                                                        GROUP BY cp.id_product ORDER BY position, id_product', false);
+                                                        GROUP BY cp.id_product ORDER BY cp.position, cp.id_product', false);
```
jocel1: with the GROUP BY, if you do not use MIN(cp.position) you don't know which position you will get if there are several time the same id_product in different cat, but with different position. The current code makes sure you always use the smallest position from all possible cats, but with your code, your using a random position.

ps_facetedsearch.php
```diff
@@ -1484,14 +1484,14 @@ public function getProductByFilters(
                         AND c.active = 1)
                     '.$price_filter_query_out.'
                     '.$query_filters_from.'
-                    WHERE 1 '.$query_filters_where.' GROUP BY cp.id_product');
+                    WHERE 1 '.$query_filters_where.' GROUP BY cp.id_product ORDER BY cp.position ASC');
```
jocel1: ORDER BY is useless, we are return a list of id, we don't care to have it ordered or not

ps_facetedsearch.php
```diff
@@ -1484,14 +1484,14 @@ public function getProductByFilters(
...
             } else {
                 $all_products_out = Db::getInstance(_PS_USE_SQL_SLAVE_)->executeS('
                     SELECT p.`id_product` id_product
                     FROM `'._DB_PREFIX_.'product` p JOIN '._DB_PREFIX_.'category_product cp USING (id_product)
                     '.$price_filter_query_out.'
                     '.$query_filters_from.'
-                    WHERE cp.`id_category` IN ('.implode(',', $categories).') '.$query_filters_where.' GROUP BY cp.id_product');
+                    WHERE cp.`id_category` IN ('.implode(',', $categories).') '.$query_filters_where.' GROUP BY cp.id_product ORDER BY cp.position ASC');
```
jocel1: same here

manfield:
Hi,
the main issue was line 1443:

`STRAIGHT_JOIN '._DB_PREFIX_.'category_product cp ON `

Let me check other points and I will be back to you

jocel1:
@manfield :`'._DB_PREFIX_.'category_product cp STRAIGHT_JOIN '._DB_PREFIX_.'category c ` AND `'._DB_PREFIX_.'category c STRAIGHT_JOIN '._DB_PREFIX_.'category_product cp ` should return the same return as it's a simple join, it only changes potentially the join speed.

manfield:
I had an issue on that query. I will come back with more details.
It has been a while I did not check that.

jocel1:
But you could check if you still have the issue with the code in #25?

manfield:
I do now

manfield:
Actually it does not work.
This is the product sorting in backend:



This is with my code:


This is with #25 code

jocel1:
what's the sorting inside the Women category ? (since it's the category you're requesting in the facetedsearch)

manfield:
Hi,
it is exactly the one I posted in the screenshot.

jocel1:
thanks a lot I've reproduced the issue locally, I'm taking a look at the code right now!

jocel1:
FYI your fix is not correctly, but the issue is when the ps_layered_full_tree mode is activate, it also checked the product order in the subcategory, and if it's lower than the current cat (I assume in your case in the TOPS & DRESSES categories, there's no position set, so the value is 0), it override the value of the current cat. We need to take into account into the query that the position of the current cat should be used in priority if the product is also present in subcategories

jocel1:
I've just pushed a fix, let me know if it works better for you

eternoendless:
Closing this one as it's fixed in https://github.com/PrestaShop/ps_facetedsearch/pull/25

PR #29, opened 2017-10-30
Creation of better ranges

costamojan:
* the artificial creation of ranges in case of an equal min and max has been dropped.
* the function mergeRanges now merges records first for entries with an equal min and max, so the user doesn't get confronted with weird ranges (previously a record [min=400, max=400] caused a range of 399..401 which is not only weird, it's confusing too as it doesn't resemble the data)
* the function mergeRanges didn't reduce the number of ranges so if the number of ranges exceeded the $outputLength it still had the same length after going through the original implementation.

Please be aware that I'm not a PHP developer, so there might be better ways to do things. Nevertheless I finally got proper ranges with these changes.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/ps_facetedsearch/29)

### DIFF
--- a/ps_facetedsearch.php
+++ b/ps_facetedsearch.php
@@ -960,14 +960,23 @@ class Ps_Facetedsearch extends Module implements WidgetInterface
                 $values[] = '('.(int) $id_product.',
 					'.(int) $currency['id_currency'].',
 					'.$id_shop.',
-					'.(int) $min_price[$currency['id_currency']].',
-					'.(int) Tools::ps_round($max_price[$currency['id_currency']] * (100 + $max_tax_rate) / 100, 0).')';
+					'.(int) Ps_Facetedsearch::taxPrice($min_price[$currency['id_currency']], $max_tax_rate).',
+					'.(int) Ps_Facetedsearch::taxPrice($max_price[$currency['id_currency']], $max_tax_rate).')';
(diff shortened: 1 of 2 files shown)


kanecko:
@costamojan I've tested it a bit and it looks much, much better than the original ranges.
However there is one bug I've found:
- I have a subcategory with 16 products
- Their prices are set at 0.90, 1.50, 1.60, 1.80, four products at 2.00, two products at 2.20, two at 2.50, 2.80, 3.80, two at 4.00
- The ranges generated are 0-1 (4 products) and 2-4 (12 products)
- The problem is that if I choose the 0-1 range, I will actually get the products with the following prices: 0.90, 1.50, 1.60, 1.80

I've only found this bug, if I have a product with a price less than 1.00.
When the range starts at, let's say, 1.00, it works just fine.

What do you suggest I do?

kanecko:
Upon further testing, it seems that the first range in the list is wrong most of the time.
I've first filtered by some attribute and then I've got two price ranges. I've tried further filtering by the first price range of 3-4 and it's showing a product with price of 4.5.

Other price ranges work fine.

PierreRambaud:
Regarding the latest big refactoring with the introduction of a slider range, this one is not needed anymore.

Kind Regards

PR #30, opened 2017-11-09
Fix notices of PHP 7.2

Quetzacoalt91:
Fixes notices about `count()` must be used on a value which implements Countable.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/ps_facetedsearch/30)

### DIFF
--- a/ps_facetedsearch.php
+++ b/ps_facetedsearch.php
@@ -2372,7 +2372,7 @@ class Ps_Facetedsearch extends Module implements WidgetInterface
 
     private static function getQuantityFilterSubQuery($filter_value, $ignore_join = false)
     {
-        if (count($filter_value) == 2 || empty($filter_value)) {
+        if (empty($filter_value) || count($filter_value) == 2) {
             return array();
         }
 
@@ -2401,7 +2401,7 @@ class Ps_Facetedsearch extends Module implements WidgetInterface
 
     private static function getConditionFilterSubQuery($filter_value, $ignore_join = false)
     {
-        if (count($filter_value) == 3 || empty($filter_value)) {
+        if (empty($filter_value) || count($filter_value) == 3) {
             return array();
         }
 


xitromedia:
Hi,
this fix is right and very urgent for developers having dev-mode on, because PHP sets a warning and PS throws a dying error making use of dev-mode impossible. The only change that should additionally be made to the fix is exchanging == with ===.

By the way: I think it is not a good idea to enable Symfony's error handling in controller class with only Debug::enable(). This causes even only a deprecated warning to throw dying error. An appropriate  error reporting level should be set here when initializing the error handling. Maybe a good idea would be a new constant beside _PS_MODE_DEV_ that holds the error reporting level and developers can modify as needed in defines.inc.php. 

Thanks for reading
André

Quetzacoalt91:
Hello @xitromedia,

Thanks for the thumb up.
About your suggestion, you have to know that the main purpose of this constant is to show absolutely everything useful for development. Displaying all notices allow us and the module developers to be rigorous on the code we write. :)

xitromedia:
hi Quetzacoalt91

yes, i know this purpose, but it avoids a theme or module develover from usefully getting debug infos  about his code when PS's code throws dyings for PS-code the developer cannot update instantly. The fix of that PS code comes only with next PS update, which may take a long time as we all know ;o). 
In my case this happened because PS is not up to date with PHP's most actual version but i have a more actual dev environment. So I think is would be more useful to not die for non-critical warnings, but only print the warning above or below the content.

André

LittleBigDev:
:lgtm:

---

Reviewed 1 of 1 files at r1.
Review status: all files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/prestashop/ps_facetedsearch/30#-:-L5K6TY7fJL5Kb3f7NXn:bnfp4nl)*

eternoendless:
Thank you @Quetzacoalt91

PR #33, opened 2018-04-11
Fix id_category NULL

darkiron:
get id FROM id not id_category (same attribute)

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/ps_facetedsearch/33)


jocel1:
the null id_category issue was actually caused by a bug in the core already fixed by https://github.com/PrestaShop/PrestaShop/pull/8962

PierreRambaud:
PR not needed anymore.
Thanks @darkiron for your contribution

PR #35, opened 2018-06-09
Fix sort by price

stickyrst:
Before, ps_facetedsearch used column p.price to sort by price, but this is not correct, because it does not take into consideration catalog rules. This commit takes the prices from ps_layered_price_index to sort by price.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/ps_facetedsearch/35)


PierreRambaud:
Really sorry to answer just now. 
We are currently working on a huge refactoring (#52) where we can maybe include what you did, but unfortunately we can't merge it in the dev branch.
Kind regards

PierreRambaud:
Your patch has been applied in #52 
Kind regards

PR #37, opened 2018-08-14
Fix hiding filters and counting of products

PatrykMakowski:
When we chose more than one value in one feature, then the hiding filters and counting of products in the block didnt work.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/ps_facetedsearch/37)


PierreRambaud:
Hey, sorry for the delay.
We are working on a huge refactoring of this module, and this bug is already fixed for the next release :)

Kind regards

PR #39, opened 2018-08-15
Dev

aixos:
Product weight filter
Hi,
he repeats it twice and it's because of the range he takes
example of a product that weighs 100
correct would be
0 - 100 (1)
101 - 200
etc..

now shows
0 - 100 (1)
100 - 200 (1)

Could it be fixed?
Thank you

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/ps_facetedsearch/39)


Quetzacoalt91:
The price ranges should be reworked in the next few months. This could be part of it.

However, note the pull-requests are supposed to suggest changes to the projects. As there is no such content to accept or decline, I have to close it.

Thanks for your understanding.

aixos:
Hi 
sorry , can you close this post
Thank you

PR #40, opened 2018-08-15
Dev

aixos:
#39
sorry
it's weight, I'm talking about weight
thank you very much.

case 'weight':
                    if ($selected_filters['weight'][0] != 0 || $selected_filters['weight'][1] != 0) {
                        $query_filters_where .= ' AND p.`weight` BETWEEN ' . (float)($selected_filters['weight'][0] - 0.001) . ' AND ' . (float)($selected_filters['weight'][0] + 0.001);

#39 
Product weight filter
Hi,
he repeats it twice and it's because of the range he takes
example of a product that weighs 100
correct would be
0 - 100 (1)
101 - 200
etc..

now shows
0 - 100 (1)
100 - 200 (1)

Could it be fixed?
Thank you

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/ps_facetedsearch/40)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -13,7 +13,8 @@
     "php": ">=5.4"
   },
   "config": {
-    "preferred-install": "dist"
+    "preferred-install": "dist",
+    "prepend-autoloader": false
   },
   "type": "prestashop-module",
   "require-dev": {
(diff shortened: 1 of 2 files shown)


Quetzacoalt91:
Sorry @aixos, but as I said in #39, opening a pull-request is not the proper way to warn about an issue.

You should open an issue here: https://github.com/PrestaShop/PrestaShop/issues

I'm closing this PR which cannot be merged.

PR #43, opened 2018-09-07
Fix [BOOM-2025] ps_facetedsearch - Filter products by the stock of matched variants, not just any of them

PeNov:
https://github.com/PrestaShop/PrestaShop/issues/10018

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/ps_facetedsearch/43)


jolelievre:
can you run php-cs-fixer please?

jolelievre:
also the version should be updated so that the module is opened to upgrade

Quetzacoalt91:
Thank you @PeNov

khouloudbelguith:
Hi @PeNov,

We have this issue https://github.com/PrestaShop/PrestaShop/issues/12143 after upgrade from v2.2.0 to v2.2.1.
Thanks!

pdkey:
@khouloudbelguith thanks for your info - i will wait for your solution.

PR #44, opened 2018-09-20
fix price / weight filter

jocel1:
Price / weight filter block should not apply there own filters (otherwise they will always disappear if we filter on price / weight because only one choice will remain)

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/ps_facetedsearch/44)


PierreRambaud:
I close this one because of the big refactoring you did.

PR #47, opened 2018-10-31
Faceted filter doesn't shows attribute selection if category have only one product with same attribute or feature.

dineshbadrukhiya:
When Category have only one product with same attribute or feature, It doesn't shows filter option with that attribute or feature on left-column. It's necessary to show filter option if at least one product have same attribute or feature.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/ps_facetedsearch/47)

### DIFF
--- a/src/Ps_FacetedsearchProductSearchProvider.php
+++ b/src/Ps_FacetedsearchProductSearchProvider.php
@@ -285,7 +285,7 @@ class Ps_FacetedsearchProductSearchProvider implements ProductSearchProviderInte
                 }
             }
             $facet->setDisplayed(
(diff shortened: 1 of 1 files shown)


dbadrukhiya:
wdyt? @Quetzacoalt91

dineshbadrukhiya:
@eternoendless Wdyt?

PierreRambaud:
Hey, thanks for your contribution. We are currently working on a big refactoring of this module.
Your pull request is not required anymore, we are sorry.

As you can see, we have category that contains one product.
image

Thanks @dineshbadrukhiya

PR #50, opened 2019-01-16
Fix issue with out of stock

dineshbadrukhiya:
When you filter color/pattern attributes, the module doesn't show results of combinations out of stock.
Fix: https://github.com/PrestaShop/PrestaShop/issues/12143

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/ps_facetedsearch/50)

### DIFF
--- a/ps_facetedsearch.php
+++ b/ps_facetedsearch.php
@@ -1420,7 +1420,7 @@ class Ps_Facetedsearch extends Module implements WidgetInterface
                         ON (pa.`id_product_attribute` = pac.`id_product_attribute`)' .
                         Shop::addSqlAssociation('product_attribute', 'pa') .
                         ' JOIN `'._DB_PREFIX_.'stock_available` sa
-                        ON (sa.`id_product_attribute` = pac.`id_product_attribute` AND sa.`quantity` > 0)
+                        ON (sa.`id_product_attribute` = pac.`id_product_attribute` AND sa.`quantity` >= 0)
                         WHERE ' . implode(' OR ', $sub_query) . ') ';
                     }
                     break;


dineshbadrukhiya:
@khouloudbelguith  is it ok?

khouloudbelguith:
Hi @dineshbadrukhiya,

Thanks for the fix.
Now your PR should be reviewed by our developers.
Thanks!

dineshbadrukhiya:
@khouloudbelguith I hope so as already one previous PR with this module is waiting for review since 3 months. I don’t know will it be worthwhile to spend time if nobody taking interest.

mwijngaard:
I just ran in to this issue as well, but this PR does not seem correct either. I think it would be better if the JOIN to the stock_available table is left as is, but NOT added to the query if stock management is disabled. If stock management is enabled, then it would be better to hide results.

dineshbadrukhiya:
@mwijngaard Yes, I agree with you. When stock management is disabled stock_available should be left.

ps_facetedsearch.php
```diff
@@ -1420,7 +1420,7 @@ public function getProductByFilters(
                         ON (pa.`id_product_attribute` = pac.`id_product_attribute`)' .
                         Shop::addSqlAssociation('product_attribute', 'pa') .
                         ' JOIN `'._DB_PREFIX_.'stock_available` sa
-                        ON (sa.`id_product_attribute` = pac.`id_product_attribute` AND sa.`quantity` > 0)
+                        ON (sa.`id_product_attribute` = pac.`id_product_attribute` AND sa.`quantity` >= 0)
```
PierreRambaud: if sa.quantity needs to be >= 0, we can remove 
```
AND sa.`quantity`
```

ping @jocel1 wdyt?

jocel1: Wonder if it could be possible to have a negative quantity here. Technically in the DB it's not an UNSIGNED so it *could* be negative

jocel1: But there are several places in the module where we are testing quantity > 0, perhaps it should be changed as well

PierreRambaud: I think it's better to take #51 which include a switch to control this kind of behavior :)

dineshbadrukhiya: Yes, @PierreRambaud  I agree with you.

PierreRambaud:
Thanks for your answer @dineshbadrukhiya 
I close this PR in favor of #51

PR #53, opened 2019-02-26
Fix Drop-down Facets being Cleared by other Drop-down Facets [BOOM-5484]

pedobry:
Using multiple dropdown facets result in random clearing of active filters, depending on order in which are filters activated.

The bug is actually in `addEncodedFacetsToFilters` method. When you have two facets(ie. Size and Brand), both as dropdown (`multipleSelectionAllowed=false`) and you select one of the filters (ie. Size), it depends in which order they are processed. When Size is processed first, it's detected as active and removed from `$activeFacetFilters`. Then Brand is processed and now it seems there is no active filter (even though there should be Size active) and it sets `setNextEncodedFacets` to wrong value.
When the facets are processed in different order (Brand before Size), Brand's `setNextEncodedFacets` is set correctly because activeFacetFilter gets deleted after (when processing Size facet).

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/ps_facetedsearch/53)

### DIFF
--- a/src/Ps_FacetedsearchProductSearchProvider.php
+++ b/src/Ps_FacetedsearchProductSearchProvider.php
@@ -231,6 +231,7 @@ class Ps_FacetedsearchProductSearchProvider implements ProductSearchProviderInte
         $urlSerializer = new URLFragmentSerializer();
 
         foreach ($facets as $facet) {
+            $facetFiltersToAdd = $activeFacetFilters;
             // If only one filter can be selected, we keep track of
             // the current active filter to disable it before generating the url stub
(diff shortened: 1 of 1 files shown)


PierreRambaud:
We are currently refactoring this module (#52 which is a really big refactoring with lot of bug fix).
Not sure your PR can be test right now.
Kind Regards,

PierreRambaud:
Because of the latest refactoring we are not able to rebase / check this PR. If the problem still happened on the dev branch, please reopen this issue and we'll investigate together :wink:

PR #57, opened 2019-05-03
Update cldr javascript library

PierreRambaud:
Update cldr library & apply cs fixer.

How to test:
Slider must works :)

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/ps_facetedsearch/57)


_dev/cldr/number-formatter.js
```diff
@@ -251,15 +252,20 @@ class CurrencyFormatter {
...
+      return formattedNumber
+        .split(CURRENCY_SYMBOL_PLACEHOLDER)
+        .join(this.numberSpecification.getCurrencySymbol());
+    }
+
+    return formattedNumber;
   }
 
   static build(specifications) {
     const symbol = new NumberSymbol(...specifications.symbol);
```
jolelievre: You can have a symbol even when it's not a price formatting?

_dev/cldr/number-formatter.js
```diff
@@ -251,15 +252,20 @@ class CurrencyFormatter {
...
-    const currency = new CurrencyFormatter(
-      new PriceSpecification(
+    let specification;
+    if (specifications.currencySymbol) {
```
jolelievre: You don't need to check if it exists? Like with `specifications.currencySymbol != undefined` or `typeof specifications.currencySymbol != undefined` ?

_dev/cldr/number-formatter.js
```diff
@@ -270,11 +276,24 @@ class CurrencyFormatter {
...
+        symbol,
+        parseInt(specifications.maxFractionDigits, 10),
+        parseInt(specifications.minFractionDigits, 10),
+        specifications.groupingUsed,
+        specifications.primaryGroupSize,
+        specifications.secondaryGroupSize,
+      );
+    }
+
+    const currency = new NumberFormatter(specification);
```
jolelievre: Maybe you should name it `currency` if it's for generic now

_dev/cldr/number-formatter.js
PierreRambaud: I want to check for a value != null && != undefined

_dev/cldr/number-formatter.js
PierreRambaud: Yep it's not a currency symbol, it's a list of symbol for the number format  https://github.com/PrestaShop/PrestaShop/blob/develop/src/Core/Localization/Specification/NumberSymbolList.php

_dev/cldr/number-formatter.js
PierreRambaud: you're right!

_dev/cldr/number-formatter.js
jolelievre: Yes what I meant is if the `currencySymbol` property is not even present in `specifications`
Does this test work?

jolelievre: Ok I tested it it works in all use cases, fine for me!

sarahdib:
@PierreRambaud 

the slider is missing 

Capture d’écran 2019-05-13 à 14 17 12

PierreRambaud:
@sarahdib My bad, forget to says this PR needs to be merged first https://github.com/PrestaShop/PrestaShop/pull/13665

PierreRambaud:
Should be ok now :)

matks:
Thank you @PierreRambaud

PR #58, opened 2019-05-10
Add phpunit tests and quality

PierreRambaud:
Add phpunit tests.

Tests: Must work as before.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/ps_facetedsearch/58)


PierreRambaud:
Can be review, more tests will come after.

ps_facetedsearch.php
```diff
@@ -93,6 +98,7 @@ public function __construct()
         $this->ps_versions_compliancy = ['min' => '1.7.1.0', 'max' => _PS_VERSION_];
 
         $this->hookDispatcher = new HookDispatcher($this);
+        $this->database = Db::getInstance();
```
Quetzacoalt91: What do you think about moving this in the first call of `getDatabase()` of this class?

PierreRambaud: Done! Thank, good idea

PR #69, opened 2019-05-23
Only rebuild table while upgrading to prevent a fatal error with memory

PierreRambaud:
This change is [Reviewable](https://reviewable.io/reviews/prestashop/ps_facetedsearch/69)


PierreRambaud:
@grabelle If field are empty, sql request returns the row with empty data, in your case, combination of idAttribute & idLang is not found this is why you have a undefined offset.

PierreRambaud:
@grabelle I revert my commit, we need to investigate more time on your issue.

PR #70, opened 2019-05-24
Reindex multistore correctly

PierreRambaud:
Tests:
- Enable multistore context
- Create a new shop
- Create templates for each store
- Reindex and clear caches.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/ps_facetedsearch/70)


ps_facetedsearch.php
```diff
@@ -265,8 +265,6 @@ public function indexAttribute($idProduct = null)
             ' . ($idProduct === null ? '' : 'AND pa.id_product = ' . (int) $idProduct) . '
             GROUP BY a.id_attribute, pa.id_product , product_attribute_shop.`id_shop`'
         );
-
-        return 1;
```
matks: Should'nt we check the index operation went well (either the database execution return or even a READ SQL query to check data looks good) ?

PierreRambaud: hehe, I planned to refactor the installation / upgrade part in the future.

matks: Easy answer, "Yes I know, I'll do it in another PR" :trollface:

PR #74, opened 2019-05-28
Render as a widget if hook is not found during the callable stack

PierreRambaud:
Fixes https://github.com/PrestaShop/PrestaShop/issues/13993

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/ps_facetedsearch/74)


marionf:
@PierreRambaud 

I hooked faceted search on displayFooterBefore
In FO when I click on "all products", _?q=Price-%24-9-35_ is added in the URL and page always loading

capture d'écran_1616

src/HookDispatcher.php
```diff
@@ -60,15 +60,24 @@ class HookDispatcher
...
+    private $module;
+
     /**
      * Init hooks
      *
      * @param Ps_Facetedsearch $module
      */
     public function __construct(Ps_Facetedsearch $module)
     {
+        $this->module = $module;
```
mickaelandrieu: I don't clearly understand what does it change to use a property instead of a local variable here: what are the side effects?

PR #76, opened 2019-05-31
Create a better indexation and add tests

PierreRambaud:
This wasn't working on faceted search 2. Now working.
Tests: You should be able to see attributes and features without editing / adding one.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/ps_facetedsearch/76)


PierreRambaud:
Tests are coming

ps_facetedsearch.php
```diff
@@ -267,6 +269,36 @@ public function indexAttributes($idProduct = null)
...
+     *
+     * @return boolean
+     */
+    public function indexFeatures()
+    {
+        return $this->getDatabase()->execute(
+            'INSERT INTO `' . _DB_PREFIX_ . 'layered_indexable_feature` ' .
+            'SELECT id_feature, 1 FROM `' . _DB_PREFIX_ . 'feature` ' .
+            'WHERE id_feature NOT IN (SELECT id_feature FROM ' .
+            '`' . _DB_PREFIX_ . 'layered_indexable_feature`)'
```
matks: Do we really need the backticks ? This would make it harder to use the module with PostgreSQL

src/Filters/Block.php
```diff
@@ -712,20 +711,24 @@ private function getAttributesBlock($filter, $selectedFilters, $idLang)
...
             if (!isset($attributesBlock[$idAttributeGroup])) {
                 $attributeGroup = $attributesGroup[$idAttributeGroup];
 
                 $attributeGroupLayeredInfos = $this->getAttributeGroupLayeredInfos($idAttributeGroup, $idLang);
-                if (empty($attributeGroupLayeredInfos)) {
-                    continue;
+                if (!empty($attributeGroupLayeredInfos)) {
+                    list($urlName, $metaTitle) = array_values($attributeGroupLayeredInfos);
+                } else {
+                    $urlName = null;
```
matks:
```suggestion
                    $urlName = $metaTitle  null;
```
Not sure if useful 😄

ps_facetedsearch.php
PierreRambaud: Since we didn't plan to make it work with pgsql, using backticks are still the good way for mysql :/

upgrade/upgrade-3.0.3.php
```diff
@@ -0,0 +1,49 @@
...
+        $object->unregisterHook($hookName);
+    }
+
+    $newHooks = [
+        'actionFeatureSave',
+        'actionFeatureValueDelete',
+        'displayFeatureValuePostProcess',
+    ];
+
+    return $object->registerHook($newHooks);
```
Quetzacoalt91: :+1:

upgrade/upgrade-3.0.3.php
```diff
@@ -0,0 +1,49 @@
...
+ * @author    PrestaShop SA <contact@prestashop.com>
+ * @copyright 2007-2019 PrestaShop SA
+ * @license   https://opensource.org/licenses/AFL-3.0 Academic Free License 3.0 (AFL-3.0)
+ * International Registered Trademark & Property of PrestaShop SA
+ */
+if (!defined('_PS_VERSION_')) {
+    exit;
+}
+
+function upgrade_module_3_0_1(Ps_Facetedsearch $object)
```
Quetzacoalt91: It seems the current version 3.0.2. You should change this file name & method for 3_0_3

PierreRambaud: 3.0.2 has already been released :/

Matt75: Instead of $object, I prefer use $module personally in my modules

PierreRambaud: ohh missreading
I will :D

PierreRambaud: Thanks guys!

PR #86, opened 2019-06-05
Do not erase filter when multiple selection is not accepted

PierreRambaud:
Fixes https://github.com/PrestaShop/PrestaShop/issues/14082

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/ps_facetedsearch/86)

### DIFF
--- a/src/Filters/Block.php
+++ b/src/Filters/Block.php
@@ -92,13 +92,13 @@ class Block
             'id_category',
             Tools::getValue('id_category_layered', Configuration::get('PS_HOME_CATEGORY'))
         );
-        $parent = new Category((int) $idParent, $idLang);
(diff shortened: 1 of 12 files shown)


src/Filters/Block.php
```diff
@@ -92,13 +92,13 @@ public function getFilterBlock(
             'id_category',
             Tools::getValue('id_category_layered', Configuration::get('PS_HOME_CATEGORY'))
         );
-        $parent = new Category((int) $idParent, $idLang);
+        $parent = new Category($idParent, $idLang);
```
matks: Wy remove the casting ?

PierreRambaud: already done on line 91

PR #89, opened 2019-06-11
fix bug indexed prices

hafkhami:
Failure to index prices in case of tax inactivity

If the "Use tax to filter price" setting is disabled, the price index will not be executed correctly and the price slider will not be displayed.

disabel_use_tax_filter_price
table_indexed_prices

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/ps_facetedsearch/89)


PierreRambaud:
Thanks @hafkhami

PR #103, opened 2019-06-21
Fix facet display when there is only one filter

PierreRambaud:
You should be able to see a facet even if you have only one filter.
See https://github.com/PrestaShop/PrestaShop/issues/9647 for more explanation.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/ps_facetedsearch/103)

### DIFF
--- a/src/Product/SearchProvider.php
+++ b/src/Product/SearchProvider.php
@@ -441,7 +441,7 @@ class SearchProvider implements FacetsRendererInterface, ProductSearchProviderIn
      * Keep facet status when it's a slider
      *
      * @param array $facets
-     * @param integer $totalProducts
+     * @param int $totalProducts
      */
     private function hideUselessFacets(array $facets, $totalProducts)
     {
@@ -467,7 +467,7 @@ class SearchProvider implements FacetsRendererInterface, ProductSearchProviderIn
                 $usefulFiltersCount > 1
                 ||
                 (
-                    /**
+                    /*
                      * There is only one fitler and the
                      * magnitude is different than the
                      * total products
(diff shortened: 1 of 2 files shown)


src/Product/SearchProvider.php
```diff
@@ -452,15 +453,29 @@ private function hideUselessFacets(array $facets)
...
+                ||
+                (
+                    /*
+                     * There is only one fitler and the
+                     * magnitude is different than the
+                     * total products
+                     */
+                    count($facet->getFilters()) === 1
+                    && $totalFacetProducts != $totalProducts
+                    && $usefulFiltersCount > 0
```
Quetzacoalt91: That's not the easier code to read. Maybe having the 2 comments at the same indentation would help?

PierreRambaud: something like.
```php
                    count($facet->getFilters()) === 1 // There is only one filter
                    && $totalFacetProducts != $totalProducts // And the magnitude is different
```

Quetzacoalt91:
```php
            $facet->setDisplayed(
                // There are two filters displayed
                $usefulFiltersCount > 1
                ||
                /*
                 * There is only one fitler and the
                 * magnitude is different than the
                 * total products
                 */
                (
                    && $totalFacetProducts != $totalProducts
                    && $usefulFiltersCount > 0
                )
            );

Quetzacoalt91: In all cases, this test seems complex to read because of the comment in the middle of the statements  :)

PierreRambaud: Thanks

PR #108, opened 2019-07-03
 attribute is selected => get combination quantity

toar86:
if attribute is selected on filters, then get combination quantity not quantity, where sa.id_product_attribute = 0

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/ps_facetedsearch/108)

### DIFF
--- a/src/Adapter/MySQL.php
+++ b/src/Adapter/MySQL.php
@@ -259,14 +259,14 @@ class MySQL extends AbstractAdapter
             'out_of_stock' => [
                 'tableName' => 'stock_available',
                 'tableAlias' => 'sa',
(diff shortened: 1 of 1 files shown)


PierreRambaud:
Can you please create an issue on https://github.com/PrestaShop/PrestaShop/issues/new/choose. I need to get the point of view of Product Managers :)

El8ctron:
При выборе различных фильтров, результат фильтрации - не вычитается, а складывается.

When choosing various filters, the filtering result is not subtracted, but added up.

PierreRambaud:
Will be fixed with https://github.com/PrestaShop/ps_facetedsearch/pull/132

PR #117, opened 2019-07-25
Remove unwanted curl

PierreRambaud:
Remove unwanted curl which is sometimes failing

Tests:
 Make sure you can install / upgrade the module :)

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/ps_facetedsearch/117)


matks:
Looks good, but why was the curl added in the 1st place ? I'm asking to be sure we dont disable some feature (because I dont see any replacement)

PierreRambaud:
This is since the first commit of this module, without comment or reason.
Since it can cause trouble for someone, and because the file_get_contents do exactly the same as the called function (indexPrices) we have no reason to keep this one.

PR #119, opened 2019-07-30
Number of product are not ok after selecting filters

PierreRambaud:
Fixes https://github.com/PrestaShop/PrestaShop/issues/14734

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/ps_facetedsearch/119)

### DIFF
--- a/src/Adapter/AbstractAdapter.php
+++ b/src/Adapter/AbstractAdapter.php
@@ -111,11 +111,8 @@ abstract class AbstractAdapter implements InterfaceAdapter
         }
 
         $operations = $this->operationsFilters->offsetGet($filterName);
-        foreach ($operations as $filterId => $filterValue) {
-            if ($filterId === $value) {
-                unset($operations[$filterId]);
-                break;
-            }
+        if (isset($operations[$value])) {
+            unset($operations[$value]);
         }
 
         if (empty($operations)) {
--- a/src/Adapter/AbstractAdapter.php
+++ b/src/Adapter/AbstractAdapter.php
@@ -100,23 +100,28 @@ abstract class AbstractAdapter implements InterfaceAdapter
      */
     public function resetOperationsFilter($filterName, $value = null)
     {
-        if ($this->operationsFilters->offsetExists($filterName)) {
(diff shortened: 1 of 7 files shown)


src/Adapter/AbstractAdapter.php
```diff
@@ -97,10 +98,25 @@ public function resetFilter($filterName)
...
      * {@inheritdoc}
      */
-    public function resetOperationsFilter($filterName)
+    public function resetOperationsFilter($filterName, $value = null)
     {
         if ($this->operationsFilters->offsetExists($filterName)) {
-            $this->operationsFilters->offsetUnset($filterName);
+            if ($value === null) {
+                $this->operationsFilters->offsetUnset($filterName);
+            } else {
```
Progi1984: You can remove the else with adding a `continue` after the `offsetUnset` call

PierreRambaud: We can't use `continnue` outside a loop or a switch :thinking: . But I refactor the function :)

src/Adapter/AbstractAdapter.php
```diff
@@ -97,10 +98,25 @@ public function resetFilter($filterName)
...
+                $operations = $this->operationsFilters->offsetGet($filterName);
+                foreach ($operations as $filterId => $filterValue) {
```
Progi1984: May be more clean

```
if (isset($operations[$value])) {
   unset($operations[$value]);
}
```

PierreRambaud: I want to check strictly on this index ^^

PierreRambaud: Oh thanks, Didn't see (came from test I think :sweat_smile: )

src/Adapter/AbstractAdapter.php
Progi1984: #fail I see a foreach :p

src/Adapter/AbstractAdapter.php
Progi1984: :ok_hand:

PR #122, opened 2019-08-13
Column type size

PrestaworksJonas:
The TEXT type column for `layered_filter_block.data` has a size limit that is too small for customers with a lot of products. I have a client that has a data size of 150kB which throws an exception since the text type is too small.

ps_facetedsearch_error

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/ps_facetedsearch/122)


PierreRambaud:
Hi and thanks for your contribution, can you please run cs-fixer? :)

PierreRambaud:
No need since a long time. I'm closing it!

PR #130, opened 2019-09-02
Add hook for Specifc Prices and move price min and max to decimal

PierreRambaud:
Fixes https://github.com/PrestaShop/PrestaShop/issues/15223

### DIFF
--- a/ps_facetedsearch.php
+++ b/ps_facetedsearch.php
@@ -67,6 +67,13 @@ class Ps_Facetedsearch extends Module implements WidgetInterface
     ];
 
     /**
+     * Number of digits for MySQL DECIMAL
+     *
+     * @var int
+     */
+    const DECIMAL_DIGITS = 5;
+
+    /**
      * @var bool
      */
     private $ajax;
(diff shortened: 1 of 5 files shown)


ps_facetedsearch.php
```diff
@@ -543,24 +550,24 @@ public function indexProductPrices($idProduct, $smart = true)
                     $minPriceValue = array_key_exists($idCountry, $minPrice) ? $minPrice[$idCountry][$currency['id_currency']] : 0;
                     $maxPriceValue = array_key_exists($idCountry, $maxPrice) ? $maxPrice[$idCountry][$currency['id_currency']] : 0;
                     if (!in_array($taxRateByCountry['iso_code'], self::ISO_CODE_TAX_FREE)) {
-                        $minPriceValue = Tools::ps_round($minPriceValue * (100 + $taxRate) / 100, 0);
-                        $maxPriceValue = Tools::ps_round($maxPriceValue * (100 + $taxRate) / 100, 0);
+                        $minPriceValue = Tools::ps_round($minPriceValue * (100 + $taxRate) / 100, self::DECIMAL_DIGITS);
```
matks: So you go for 5 digits instead of 0 ?

PierreRambaud: Yep because of rounding.
Example:
 - Red scarf 10.60€
 - Blue scarf 10.80€
 - Yellow scarf 10.70€

The previous behaviour was saving 11€ for each products, which means you're not able to properly request and check the price order :/
I use 5 because digits will be configurable in the BO, and can be higher than 2.

PR #131, opened 2019-09-03
UI/UX improvement - scroll to #products container on updateProductList event

PrestaShark:
Scroll to #products container on updateProductList event.

Works fine as standalone feature - probably need to be fixed in this code (im not jq expert - just shark ;-) )

Cheers!

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/ps_facetedsearch/131)

### DIFF
--- a/views/dist/front.js
+++ b/views/dist/front.js
@@ -182,5 +182,5 @@ var S=n(3),x=function(){function t(e){!function(t,e){if(!(t instanceof e))throw 
  * @license   https://opensource.org/licenses/AFL-3.0 Academic Free License 3.0 (AFL-3.0)
  * International Registered Trademark & Property of PrestaShop SA
  */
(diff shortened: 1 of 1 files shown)


PrestaShark:
One flaw. Might be anoying when user want to select few combinations at once (like color, size). In that case when user check desired combinations then search button should be available to fire up search event.

PierreRambaud:
@PrestaShark I think this is one reason why they told me to let the theme do this job.
More because the page can be changed in different way (buttons, nav, links, ..)
Not sure this one can be integrate directly in the module.

Wdyt @PrestaShark 

(If yes, the PR'll be updated because we can't update assets manually :D )

marionf:
Hello @PrestaShark 

As you can see here, we already talked about this here: https://github.com/PrestaShop/PrestaShop/issues/10993
https://github.com/PrestaShop/PrestaShop/issues/14740

And we don't want to change this behavior for the reasons explained in these issues:

> This is something each individual shop owner and / or theme should be in control of and not the core platform nor facetedsearch module

PR #132, opened 2019-09-04
Calculate the number of available products with the stock management

PierreRambaud:
Fixes: https://github.com/PrestaShop/PrestaShop/issues/15056

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/ps_facetedsearch/132)


src/Product/SearchProvider.php
```diff
@@ -456,7 +456,7 @@ private function hideUselessFacets(array $facets, $totalProducts)
                 (
                     count($facet->getFilters()) === 1
                     && $totalFacetProducts < $totalProducts
-                    && $usefulFiltersCount > 1
+                    && $usefulFiltersCount > 0
```
matks: This can lead to massive changes in the behavior, sure about it ?

PierreRambaud: image
As you can see in this image, you have only one `Color` which is not `Not available`, without the change, it doesn't display it.
More, because the facet doesn't have the same number of products than the total, it means the filter is useful and can be used.

matrixino:
@PierreRambaud I'm still getting products that shouldn't show after applying this diff.
If I set for example Size 43 in a shoe shop, I'm still getting some products which don't have size 43 as an option.
Does this only works if the product has the size 43 combination set at 0 qty?
Is it possible to hide all the product which don't have that combination size at all, without having to add that size/combination to products that will never have it?

marionf:
@matrixino This is another improvement request, please follow this issue: https://github.com/PrestaShop/PrestaShop/issues/14518#issuecomment-518119334

matrixino:
> @matrixino This is another improvement request, please follow this issue: [PrestaShop/PrestaShop#14518 (comment)](https://github.com/PrestaShop/PrestaShop/issues/14518#issuecomment-518119334)

Thanks, I fixed it myself

PR #138, opened 2019-09-13
Calculate the number of available products with the stock management

PierreRambaud:


PierreRambaud:
Already validated here: https://github.com/PrestaShop/ps_facetedsearch/pull/132

PR #144, opened 2019-09-25
Use public_name instead of name for attribute group

PierreRambaud:
Fixes https://github.com/PrestaShop/PrestaShop/issues/15688

It should use the Public name instead of Name property for attributes.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/ps_facetedsearch/144)

### DIFF
--- a/src/Filters/Block.php
+++ b/src/Filters/Block.php
@@ -678,7 +678,7 @@ class Block
         if (!isset($this->attributesGroup[$idLang])) {
             $this->attributesGroup[$idLang] = [];
             $tempAttributesGroup = $this->database->executeS(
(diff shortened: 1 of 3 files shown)


Robin-Fischer-PS:
Hi @PierreRambaud , thanks for the PR :)

I've got a problem with it:
When "Name" and "Public name" are identical, the filter works well.
But when "Name" and "Public name" are different, the filter doesn't work anymore : When clicking on a choice, the filter is not added.

See video below :

https://drive.google.com/file/d/1R3MM0E94RbDjg_cXup-ugU-6GR8CgnYw/view?usp=sharing

Thanks :)

PierreRambaud:
@Robin-Fischer-PS Thanks! Should be ok now!

matks:
Thank you @PierreRambaud

PR #151, opened 2019-12-05
Fix CLDR number formatter for negative values

jolelievre:
The component tried to detect if the number was negative on the wrong variable (absolute one so always positive)

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/ps_facetedsearch/151)


PierreRambaud:
No need QA, Thanks @jolelievre

PR #156, opened 2020-01-22
search products between two features values

gmisterk:
Hi all
I have create two features (age_min and age_max) and i want to apprly this critera in sql request.
My target is to find all product between age_min and age_max values

i think that i need to change  ps_facetedsearch.php but i Don't know where exactly must i change sql request.

anybody can help me ?

Regards
PS 1.7

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/ps_facetedsearch/156)


gadnis:
@gmisterk file MySQL.php responsible for sql query. Maybe there you will find what you are looking for. Also check tests folder, those files are like a documentation.

PierreRambaud:
Since there is no commit in this pull request, I'm closing it.
Please create an issue https://github.com/PrestaShop/PrestaShop/issues/new/choose if you have any problems!
Kind regards

PR #157, opened 2020-02-17
Fix order by price indexation without tax rate

clotaire202:
If you have several shop with different default country, indexation of price are base only on the default country of the default shop.
For instance, you can create a shop in France and a shop in USA with only one currency.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/ps_facetedsearch/157)

### DIFF
--- a/ps_facetedsearch.php
+++ b/ps_facetedsearch.php
@@ -411,7 +411,7 @@ class Ps_Facetedsearch extends Module implements WidgetInterface
             );
 
             if (empty($taxRatesByCountry) || !Configuration::get('PS_LAYERED_FILTER_PRICE_USETAX')) {
-                $idCountry = (int) Configuration::get('PS_COUNTRY_DEFAULT');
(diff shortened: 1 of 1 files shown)


PierreRambaud:
With your change, it takes only the default shop, but this isn't what we want :thinking:

202-ecommerce:
No because this code is in a foreach loop of all shops.
Get configuration values with shop id return the DEFAULT_COUNTRY of the current shop if it’s define or the default value from all shop context (shop_id null on ps_configuration table).

PierreRambaud:
@202-ecommerce haha you're right, I got confused with the red and green line -_-'

PierreRambaud:
No need QA, thanks @clotaire202 and @202-ecommerce (Sometimes I'm a little bit stupid -_-)

matks:
> No need QA, thanks @clotaire202 and @202-ecommerce (Sometimes I'm a little bit stupid -_-)

PR #158, opened 2020-02-21
Show products only from default category - optional

gadnis:
I know that there is an issue open in main prestashop git repo and there is a lot of information and opinions 

I believe merchant must be able to choose to turn off the "show only from default category" option when not using "Show products from subcategories" option. 

Why? 
We have complex category tree in our shop. There is category for summer discounts, b-stock, some other offer categories. So my product must be in several categories. I don't want to show products in each category like "Show products from subcategories" option does, it is to messy when having thousand of products.

Please add this option or tell me why not 😁

Screen Shot 2020-02-22 at 19 45 49

Issue url: [https://github.com/PrestaShop/PrestaShop/issues/14003](https://github.com/PrestaShop/PrestaShop/issues/14003)

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/ps_facetedsearch/158)


SD1982:
This PR correctly fix [#14003](https://github.com/PrestaShop/PrestaShop/issues/14003). Thanks @gadnis

Progi1984:
Thanks @gadnis

PR #159, opened 2020-02-24
Fix price indexing for products with specific prices only for some currencies

Perfecticus:
Hi, I think the condition to skip default prices is too wide here.
I have a shop where I define base prices in default currency.
Prices for other currencies are defined using specific prices, in order to avoid rounding issues due to exchange rates.

The price calculation for such a product then is:
default currency: default price
other currencies: specific price.

The logic for indexing a products price in this file looks for ANY specific price rule, regardless of id_currency/id_country for the product, and if it finds one, the default price is never used.

As a result in my price index, for default currency (no specific rule) all products are indexed with zero price.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/ps_facetedsearch/159)

### DIFF
--- a/ps_facetedsearch.php
+++ b/ps_facetedsearch.php
@@ -428,12 +428,6 @@ class Ps_Facetedsearch extends Module implements WidgetInterface
 
                 // Get price by currency & country, without reduction!
                 foreach ($currencyList as $currency) {
-                    if (!empty($productMinPrices)) {
-                        $minPrice[$idCountry][$currency['id_currency']] = null;
-                        $maxPrice[$idCountry][$currency['id_currency']] = null;
-                        continue;
-                    }
-
                     $price = Product::priceCalculation(
                         $idShop,
                         (int) $idProduct,


Perfecticus:
Example

Perfecticus:
Hmm no, I believe the problem was introduced in commit e1e3b70.

While I understand the issue with indexing all combinations of specific prices here, at the very least the base price (non-specific) should be indexed if no specific price exist for the current conditions.

PierreRambaud:
Let's have a try, hope the QA team will find no regression :)

sarahdib:
@Perfecticus Could you provide us a scenario in order to test this PR ?

Perfecticus:
Make a shop with 2 currencies.
Create a product with a non-zero default price.
Apply a non-zero specific price with currency condition for one of the currencies.

Run indexing a find that for the currency which does not have a specific price, the price will be zero in the price index table as shown in the screenshot above.

matthieu-rolland:
Thank you @Perfecticus !

PR #170, opened 2020-05-27
Ajax update product list on select lists

Leenwt:
Added  js-search-link to update the list via ajax instead of page reload on a dropdown selection.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/ps_facetedsearch/170)


Progi1984:
Thanks @Leenwt & @sarahdib

PR #175, opened 2020-06-15
Fix license headers

PierreRambaud:
This change is [Reviewable](https://reviewable.io/reviews/prestashop/ps_facetedsearch/175)


composer.json
```diff
@@ -1,46 +1,49 @@
...
-  "require-dev": {
-    "friendsofphp/php-cs-fixer": "^2.14",
-    "phpunit/phpunit": "~5.7",
-    "prestashop/php-coding-standards": "dev-master",
-    "mockery/mockery": "^1.2"
-  },
-  "config": {
-    "platform": {
-      "php": "5.6.0"
+    "name": "prestashop/ps_facetedsearch",
```
matks, line 2: Indented but I think it's harmess

PierreRambaud: I even don't know how the indentation was changed (I only used header-stamp :facepalm: )

matks: It's a known bug of header-stamp. It decodes the entire JSON, adds its stamp, then use `file_put_contents(json-encode($content));` to put content back into composer.json in so the indent used will be the one used on the machine (dont know how it's chosen).

PierreRambaud: :facepalm:

PR #189, opened 2020-08-03
Fix order by price indexation without tax rate (default country fix)

denys202:
| Questions     | Answers
| Description?  | When you configure default sort by for category page with value `sort by price` and products in category don't have a tax configured (without tax) or an option `use taxes` is disabled in the module and you have more than one country configured for current shop, so in this case you will not see products in category. It happens only when you access FO using customer that country is different from the default country. The module tries to make joins by current country id, but price index use only default country for each shop/product (the line is not present in DB, because we made indexes only for default country). To fix this problem i propose to fetch activated countries for each shop and add them all to the table of price indexes (not only default country).
| Type?         | bug fix
| BC breaks?    | no
| Deprecations? |  no
| Fixed ticket? | Fixes #157.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/ps_facetedsearch/189)

### DIFF
--- a/ps_facetedsearch.php
+++ b/ps_facetedsearch.php
@@ -413,13 +413,13 @@ class Ps_Facetedsearch extends Module implements WidgetInterface
             if (empty($taxRatesByCountry) || !Configuration::get('PS_LAYERED_FILTER_PRICE_USETAX')) {
                 $shopCountries = Country::getCountriesByIdShop($idShop, $this->getContext()->language->id);
                 $taxCountries = array_filter($shopCountries, function ($country) {
-                    return $country['active'] == 1;
+                    return $country['active'];
                 });
                 $taxRatesByCountry = array_map(function ($country) {
                     return [
                         'rate' => 0,
                         'id_country' => $country['id_country'],
-                        'iso_code' => $country['iso_code']
+                        'iso_code' => $country['iso_code'],
                     ];
                 }, $taxCountries);
             }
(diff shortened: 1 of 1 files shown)


ps_facetedsearch.php
```diff
@@ -411,9 +411,17 @@ public function indexProductPrices($idProduct, $smart = true)
             );
 
             if (empty($taxRatesByCountry) || !Configuration::get('PS_LAYERED_FILTER_PRICE_USETAX')) {
-                $idCountry = (int) Configuration::get('PS_COUNTRY_DEFAULT', null, null, $idShop);
-                $isoCode = Country::getIsoById($idCountry);
-                $taxRatesByCountry = [['rate' => 0, 'id_country' => $idCountry, 'iso_code' => $isoCode]];
+                $shopCountries = Country::getCountriesByIdShop($idShop, $this->getContext()->language->id);
+                $taxCountries = array_filter($shopCountries, function ($country) {
+                    return $country['active'] == 1;
```
PierreRambaud, line 416:
```suggestion
                    return $country['active'];
```
The comparison with `==` are useless

sarahdib:
@denys202 thank you for your contribution.

The problem is not solved : https://drive.google.com/file/d/1BEPXNM5uXES6FS_rtypa-DYjdxUu-WOf/view
Could you check the video and tell me if there is another step ?

denys202:
Hello, @sarahdib , yes, to reproduce my issue you should apply this configuration or don't apply taxes for some products (or there is one more reason, i will describe it later): 
image
After that, it will be good to rebuild index of prices and then you will be able to filter products with my fix.
But on your video, I see that there is no filtering by price. I tried to reproduce your issue, by debugging the code and i did it only once: when the product's tax rule group was not associated with selected country. So, i just added my country in tax rule group and products appeared while filtering.
Do not hesitate to write to me if you have any other questions.

sarahdib:
@denys202 thank you for the information it's OK now :)

PR #238, opened 2020-10-19
Update Converter.php

Hirashi7:
| Questions     | Answers
| Description?  | In the current version, if the attribute is set to a color, the texture will not be passed at all (this creates problems on the front-end). If we separate these conditions, both information will be available from template.
| Type?         | bug fix
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | 
| How to test?  | facetes object will have both color and texture now

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/ps_facetedsearch/238)

### DIFF
--- a/src/Filters/Converter.php
+++ b/src/Filters/Converter.php
@@ -122,7 +122,8 @@ class Converter
                         if (isset($filterArray['color'])) {
                             if ($filterArray['color'] != '') {
                                 $filter->setProperty('color', $filterArray['color']);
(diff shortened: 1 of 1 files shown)


PierreRambaud:
Hi and thanks for your contribution. Could you please create an issue on https://github.com/PrestaShop/PrestaShop?

PierreRambaud:
Hi @Hirashi7 ,

Since we had no news from you for more than 30 days, I'll close this pull request. Feel free to reopen or open another one if you think it's still relevant.

Thanks!

PR #275, opened 2020-12-04
Correctly serialize and unserialize parameters in URLs.

PierreRambaud:
| Questions     | Answers
| Description?  | Correctly serialize and unserialize parameters in URLs.
| Type?         | bug fix
| BC breaks?    | yes
| Deprecations? | yes
| Fixed ticket? | Fixes PrestaShop/Prestashop#16320
| How to test?  | See ticket.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/ps_facetedsearch/275)

### DIFF
--- a/src/Filters/Converter.php
+++ b/src/Filters/Converter.php
@@ -31,7 +31,7 @@ use Manufacturer;
 use PrestaShop\PrestaShop\Core\Product\Search\Facet;
 use PrestaShop\PrestaShop\Core\Product\Search\Filter;
 use PrestaShop\PrestaShop\Core\Product\Search\ProductSearchQuery;
-use PrestaShop\PrestaShop\Core\Product\Search\URLFragmentSerializer;
+use PrestaShop\Module\FacetedSearch\URLSerializer;
 use Tools;
 
 class Converter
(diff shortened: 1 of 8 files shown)


NeOMakinG:
@PierreRambaud PHPStan isn't happy!

PierreRambaud:
> @PierreRambaud PHPStan isn't happy!

I notice phpstan is badly implemented in GA. My fault, I'm fixing it in this pull request.

PierreRambaud:
Need to merge #276 first, then the PHPStan configuration will be fixed :)

Progi1984 (dismissed):

matks (commented):
Code is good but I think it would be not too hard to preserve backward compatibility. Is there a reason for the BC break ?

PierreRambaud:
> Code is good but I think it would be not too hard to preserve backward compatibility. Is there a reason for the BC break ?

Because it's totally broken :sweat:

Progi1984 (approved):

matks:
> > Code is good but I think it would be not too hard to preserve backward compatibility. Is there a reason for the BC break ?
> 
> Because it's totally broken 😓

Everything in PrestaShop is broken :trollface: so we can make BC everywhere?

(just a joke, I 👍  the PR)

matks (approved):

PR #295, opened 2020-12-21
Always sort by id_product when there is similar result

PierreRambaud:
| Questions     | Answers
| Description?  | When sorting products and they have the same price or the same ORDER BY in the database and this one has cache disable, the sort is random as it's explained in the issue. To be sure it never occurred, always add a second sort by id_product DESC
| Type?         | bug fix
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes PrestaShop/Prestashop#18674
| How to test?  | Turn off the `query_cache_type` in your MySQL session (show cache values with `SHOW VARIABLES LIKE '%query_cache%'`)

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/ps_facetedsearch/295)

### DIFF
--- a/src/Adapter/MySQL.php
+++ b/src/Adapter/MySQL.php
@@ -128,6 +128,7 @@ class MySQL extends AbstractAdapter
 
         if ($orderField) {
             $query .= ' ORDER BY ' . $orderField . ' ' . strtoupper($this->getOrderDirection());
+            $query .= ', p.id_product DESC';
         }
 
         if ($this->limit !== null) {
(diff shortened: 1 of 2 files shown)


atomiix (approved):

khouloudbelguith:
Hi @PierreRambaud,

It is ok :heavy_check_mark: 
I checked with PS1769, 177x branch & develop branch.
image
**BO**
image
**Without PR**
image
**With PR**
image
image

Thank you!

Progi1984 (approved):

Progi1984:
Thanks @PierreRambaud & @khouloudbelguith

PR #299, opened 2020-12-30
Cast value when saving prices

PierreRambaud:
| Questions     | Answers
| Description?  | Make sure prices are cast into int.
| Type?         | bug fix
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes PrestaShop/Prestashop#22622
| How to test?  | see issue.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/ps_facetedsearch/299)

### DIFF
--- a/ps_facetedsearch.php
+++ b/ps_facetedsearch.php
@@ -561,8 +561,8 @@ class Ps_Facetedsearch extends Module implements WidgetInterface
                     $values[] = '(' . (int) $idProduct . ',
                         ' . (int) $currency['id_currency'] . ',
                         ' . $idShop . ',
-                        ' . (int) $minPriceValue . ',
-                        ' . (int) $maxPriceValue . ',
+                        ' . (float) $minPriceValue . ',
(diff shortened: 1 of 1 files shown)


ps_facetedsearch.php
```diff
@@ -561,8 +561,8 @@ public function indexProductPrices($idProduct, $smart = true)
                     $values[] = '(' . (int) $idProduct . ',
                         ' . (int) $currency['id_currency'] . ',
                         ' . $idShop . ',
-                        ' . $minPriceValue . ',
-                        ' . $maxPriceValue . ',
+                        ' . (int) $minPriceValue . ',
+                        ' . (int) $maxPriceValue . ',
```
Seb33300, line 565: Prices can have decimals, should be `(float)`?

PierreRambaud: You're totally right :sweat_smile:

Seb33300 (approved):

NeOMakinG (approved):

PR #315, opened 2021-02-02
Hide or display others filters

AvenueDuWeb:
| Questions     | Answers
| Description?  | Possibility to display or hide others filters when only one result with active filter
| Type?         | new feature
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes partially PrestaShop/Prestashop#10018
| How to test?  | Hiding (by default) does not change the previous behaviour.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/ps_facetedsearch/315)

### DIFF
--- a/src/Product/SearchProvider.php
+++ b/src/Product/SearchProvider.php
@@ -448,7 +448,8 @@ class SearchProvider implements FacetsRendererInterface, ProductSearchProviderIn
         foreach ($facets as $facet) {
             if ($facet->getWidgetType() === 'slider') {
                 $facet->setDisplayed(
-                    $facet->getProperty('min') != $facet->getProperty('max')
+                    !(int) Configuration::get('PS_LAYERED_HIDE_OTHERS_FILTERS')
+                    || $facet->getProperty('min') != $facet->getProperty('max')
                 );
                 continue;
             }
--- a/src/Product/SearchProvider.php
+++ b/src/Product/SearchProvider.php
@@ -464,7 +464,7 @@ class SearchProvider implements FacetsRendererInterface, ProductSearchProviderIn
 
             $facet->setDisplayed(
                 // There are two filters displayed
-                $usefulFiltersCount > Configuration::get('PS_LAYERED_HIDE_OTHERS_FILTERS') ? 1 : 0
+                $usefulFiltersCount > (int) Configuration::get('PS_LAYERED_HIDE_OTHERS_FILTERS')
                 ||
                 /*
                  * There is only one fitler and the
--- a/src/Product/SearchProvider.php
+++ b/src/Product/SearchProvider.php
@@ -186,7 +186,7 @@ class SearchProvider implements FacetsRendererInterface, ProductSearchProviderIn
 
         $this->labelRangeFilters($facets);
         $this->addEncodedFacetsToFilters($facets);
-        $this->hideUselessFacets($facets, (int)$result->getTotalProductsCount());
+        $this->hideUselessFacets($facets, (int) $result->getTotalProductsCount());
 
         $facetCollection = new FacetCollection();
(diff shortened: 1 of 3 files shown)


src/Product/SearchProvider.php
```diff
@@ -464,7 +464,7 @@ private function hideUselessFacets(array $facets, $totalProducts)
 
             $facet->setDisplayed(
                 // There are two filters displayed
-                $usefulFiltersCount > 1
+                $usefulFiltersCount > Configuration::get('PS_LAYERED_HIDE_OTHERS_FILTERS') ? 1 : 0
```
PierreRambaud, line 467:
```suggestion
                $usefulFiltersCount > (int) Configuration::get('PS_LAYERED_HIDE_OTHERS_FILTERS')
```

PierreRambaud: But I'm not sure about this condition. Why not skipping this method?

AvenueDuWeb: Thanks for the suggestion for code improvement. Passing this method is problematic when you have a filter on a page that does not contain any matching products. Ex: Category when you are in a category that has no children.

PierreRambaud (dismissed):

src/Product/SearchProvider.php
```diff
@@ -186,7 +186,7 @@ public function runQuery(
 
         $this->labelRangeFilters($facets);
         $this->addEncodedFacetsToFilters($facets);
-        $this->hideUselessFacets($facets, (int) $result->getTotalProductsCount());
+        $this->hideUselessFacets($facets, (int)$result->getTotalProductsCount());
```
AvenueDuWeb:
```suggestion
        if(Configuration::get('PS_LAYERED_HIDE_OTHERS_FILTERS')) {
            $this->hideUselessFacets($facets, (int)$result->getTotalProductsCount());
        }
```

@PierreRambaud I have a problem that I hadn't identified before. It doesn't work with sliders (price, weight). So I suggest this modification which works all the time and to put the initial code back on the hideUselessFacets() function. Ok?

PierreRambaud: I think the best way is to complete the issue I linked in your PR description.
It's about facets filters that don't appear because there is only 1 element.

AvenueDuWeb: My suggestion also no longer works in the context of a filter that doesn't need to exist, e.g. a category where there would only be one product, this is similar to what I said about your answer about skipping this method. The idea is to keep the filter displayed, only if in the unfiltered list there is more than one product. It is necessary to make an additional condition on the slider. I will make a new commit for this.

PierreRambaud (approved):

src/Product/SearchProvider.php
Hlavtox: @PierreRambaud good luck to anyone understanding this after some time :D `$usefulFiltersCount > (int) Configuration::get('PS_LAYERED_HIDE_OTHERS_FILTERS')`

PierreRambaud: I will improve it later for a better understanding ;)

khouloudbelguith:
Hi @AvenueDuWeb,

I couldn't find any difference with or without PR with PS1772.
Steps to reproduce the issue:

1. Go to BO > Catalog > Products
2. Edit the product `Hummingbird printed t-shirt`
3. Set all combination quantity to 0 except the Large-black one
image
5. Go to FO > all products
6. Check Clothes in Categories
7. Check Small as Size and White as Color
8. See error => the Hummingbird printed t-shirt is displayed

It shouldn't because the product is out of stock and the option `Show unavailable, out of stock last` is disabled
image

Thanks to check and feedback.

matks:
Hello @AvenueDuWeb do you think you will continue?

PR #352, opened 2021-02-27
Set default check on filter by stock if "only show available" active

Guisardo:
| Questions     | Answers
| Description?  | If filter by stock is not set and only show available at the shop is set, then apply a default to filter only content with stock at the search results
| Type?         | improvement
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes partially PrestaShop/PrestaShop#10018
| How to test?  | At BO: At the facetedsearch config, set "Show unavailable, out of stock last" to NO; setup a few products with combinations; add stock only to some of those combinations in one product and add stock to the rest of combinations in other products   At FO: use the faceted search to filter by a particular combination that only one of the products has stock --- You should see only the products wirh stock on that filtered combination.

### DIFF
--- a/src/Filters/Converter.php
+++ b/src/Filters/Converter.php
@@ -359,6 +359,9 @@ class Converter
                     }
             }
         }
+        if (0 == Configuration::get('PS_DISP_UNAVAILABLE_ATTR') && !isset($searchFilters[self::TYPE_QUANTITY])) {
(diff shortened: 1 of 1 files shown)


PierreRambaud:
I think you're trying to do something that already in progress: #315
I'm sorry but I'm closing your pull request because work is already in progress

PR #412, opened 2021-04-29
Replace Tools::encrypt wich is decprecated since 1.7.0 by Tools::hash.

Sinepel:
| Questions     | Answers
| Description?  | Replace the deprecated Tools::encrypt function by Tools::hash. Tools::hash is here since 1.7.0 and this new version of ps_facetedsearch is compatible with version 1.7.1 or higher, so it's OK I think !
| Type?         | refacto
| BC breaks?    | no
| Deprecations? | no
| How to test?  | Check that CRON tasks do not trigger a Bad Token error due to a bad token hash

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/ps_facetedsearch/412)


PierreRambaud (approved):

PierreRambaud:
Thank you @Sinepel

PR #449, opened 2021-06-10
Fix for filter on subcategories with the same name

Prestaworks:
This pull request fix the issue with the ps_faceted search modules sub category filter. It is mixing what category it should fetch products from when there are several categories with the same name.

| Questions     | Answers
| Description?  | This pull request fix the issue with the sub category filter mixing what category it should fetch products from.
| Type?         | bug fix
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes PrestaShop/Prestashop#24764.
| How to test?  | Follow instructions in Ticket.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/ps_facetedsearch/449)

### DIFF
--- a/src/Filters/Converter.php
+++ b/src/Filters/Converter.php
@@ -389,7 +389,7 @@ class Converter
                 case self::TYPE_CATEGORY:
                     if (isset($facetAndFiltersLabels[$filterLabel])) {
                         foreach ($facetAndFiltersLabels[$filterLabel] as $queryFilter) {
-                            $categories = Category::searchByName($idLang, $queryFilter, true);
+                            $categories = Category::searchByNameAndParentCategoryId($idLang, $queryFilter, (int) $query->getIdCategory());
                             if ($categories) {
                                 $searchFilters[$filter['type']][] = $categories['id_category'];
                             }


src/Filters/Converter.php
```diff
@@ -389,7 +389,7 @@ public function createFacetedSearchFiltersFromQuery(ProductSearchQuery $query)
                 case self::TYPE_CATEGORY:
                     if (isset($facetAndFiltersLabels[$filterLabel])) {
                         foreach ($facetAndFiltersLabels[$filterLabel] as $queryFilter) {
-                            $categories = Category::searchByName($idLang, $queryFilter, true);
+                            $categories = Category::searchByNameAndParentCategoryId($idLang, $queryFilter, (int) $query->getIdCategory());
```
PierreRambaud, line 392: Is it ok when using the "All products" page?

Prestaworks: No it does not work on "all products" if the leveldepth for sub categories makes all the duplicate names categories show up in the list.
But to solve that, the "encodedFacets" have to be changed from "type-name" to "type-id". That would proboboly affect SEO?

PierreRambaud: Yes, that's the problem, I was against using type name, but it's how it started :(

PierreRambaud: At least it fixes one use case, we can improve later :P

PierreRambaud (approved):

cyrius (commented):
> Reviewed 1 of 1 files at r1.
> *[Reviewable](https://reviewable.io/reviews/prestashop/ps_facetedsearch/449)* status: all files reviewed, 1 unresolved discussion (waiting on @Prestaworks)

Robin-Fischer-PS:
Hi @Prestaworks !

Your fix is OK, and the facetedsearch still works fine : It's QA :heavy_check_mark: , thanks !

matks:
Thank you @Prestaworks and @Robin-Fischer-PS

uruk-hai-21:
Doesn't work for me
PS 1.7.7.5 - 1.7.7.3
ps_facetedsearch 3.7.1

PR #479, opened 2021-07-28
Enable filtering by stock properly

Hlavtox:
| Questions     | Answers
| Description?  | See below
| Type?         | bug fix / improvement
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes PrestaShop/Prestashop#14518
| How to test?  | See below

### Description
- This PR enables to filter products by quantity in stock, like it was in PrestaShop 1.6. You now have three options:
  - Not available - shows product with zero quantity and disabled out-of-stock ordering. (not very useful but kept it - users can remove it in their TPL)
  - Available - shows products in stock or with enabled out-of-stock ordering
  - In stock - shows products in stock (new option I added)

### How to test
1. Make 4 products
  A. Quantity 0, disabled backorder.
  B. Quantity 0, enabled backorder.
  C. Quantity 1.
  D. Product with 2 combinations, one combination quantity 1, one combination quantity 0, backorder disabled.
2. Setup filter, clear cache if needed
3. Go try the filter and see that it works properly.

You should get
- In stock - 2 - C, D
- Available - 1 - B
- Not available - 2 - A, D

### Screenshot
in stock

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/ps_facetedsearch/479)

ping @kpodemski @PierreRambaud

### DIFF
--- a/src/Filters/Block.php
+++ b/src/Filters/Block.php
@@ -410,7 +410,7 @@ class Block
     private function getQuantitiesBlock($filter, $selectedFilters)
     {
         $filteredSearchAdapter = $this->searchAdapter->getFilteredSearchAdapter(Search::STOCK_MANAGEMENT_FILTER);
-        $quantityArray = [
+        $availabilityOptions = [
             0 => [
                 'name' => $this->context->getTranslator()->trans(
                     'Not available',
@@ -420,6 +420,14 @@ class Block
                 'nbr' => 0,
             ],
             1 => [
+                'name' => $this->context->getTranslator()->trans(
+                    'Available',
+                    [],
+                    'Modules.Facetedsearch.Shop'
+                ),
+                'nbr' => 0,
+            ],
+            2 => [
                 'name' => $this->context->getTranslator()->trans(
                     'In stock',
                     [],
@@ -438,7 +446,7 @@ class Block
         }
 
         if ($this->psStockManagement) {
-            $results = [];
+            // Products without quantity in stock, with out-of-stock ordering disabled
             $filteredSearchAdapter->addOperationsFilter(
                 Search::STOCK_MANAGEMENT_FILTER,
                 [
@@ -448,10 +456,9 @@ class Block
                     ],
                 ]
             );
-            $results[0] = [
-                'c' => $filteredSearchAdapter->count(),
-            ];
+            $availabilityOptions[0]['nbr'] = $filteredSearchAdapter->count();
 
+            // Products in stock, or with out-of-stock ordering enabled
             $filteredSearchAdapter->addOperationsFilter(
                 Search::STOCK_MANAGEMENT_FILTER,
                 [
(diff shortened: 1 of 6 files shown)


src/Filters/Block.php
```diff
@@ -420,6 +420,14 @@ private function getQuantitiesBlock($filter, $selectedFilters)
                 'nbr' => 0,
             ],
             1 => [
+                'name' => $this->context->getTranslator()->trans(
```
PierreRambaud, line 423: This option should be available only when the stock management is activated, no?

Hlavtox: If you disable stock management, merchant cannot update quantity nor set out-of-stock. So the quantity block is not needed at all.

It worked before, but I modified it a bit. Maybe it will be some miliseconds faster this way.

PierreRambaud:
> @PierreRambaud Can you help me get the tests working?

I don't know if I will be able to have time to do it right now :(

PierreRambaud:
> @PierreRambaud Can you help me get the tests working? pray

I'll try to do it this week

PierreRambaud:
@Hlavtox it should be better right now, but I remove an unwanted unset, it's up to the view to display it or not.

matks:
@Hlavtox this might be a very good moment to level up your developer skills and add these tests 😄  ? adding test on a library that already have tests is easier than starting from a white page

Hlavtox:
@matks @PierreRambaud Guys, maybe some next time. 🙏 I have a lot of things to do and doing some PRs from time to time is already taking big time. 

I would spend 4 hours adding one test and it still would not work, while TEST BOSS @boubkerbribri will have it in 5 minutes :D

PierreRambaud (approved):

florine2623:
Hello @Hlavtox ,

Thanks for the PR !

With the PR, I have an issue with the pagination.
For example, in my case, I'm choosing option "Availability : In stock (17)".
Screenshot 2021-12-23 at 11 26 14
I should have 17 products shown, only 12 are displayed :
Screenshot 2021-12-23 at 11 26 19

I'm having a hard time testing with a lot of products ^^. Could you take a look please ?
Thanks!

matthieu-rolland (commented):
removed the `waiting for QA` label since QA has been done and this PR is now waiting for a fix

hibatallahAouadni:
Hello @Hlavtox 

The pagination is not displayed :x: as @florine2623 mentioned above. 
Please check the attached screen record below:

https://watch.screencastify.com/v/5FBUh7KTBeXm2ZmiYyQC

Thanks!

PierreRambaud (approved):

Hlavtox:
Readyyyyyy for final QA! :-)

PierreRambaud (approved):

florine2623:
Hello again @Hlavtox ,

Thanks for all the updates, pagination is now displayed correctly ✅ 

Tested these 4 options :
> A. Quantity 0, disabled backorder.
B. Quantity 0, enabled backorder.
C. Quantity 1.
D. Product with 2 combinations, one combination quantity 1, one combination quantity 0, backorder disabled.

The result works as intended.
- all options are independent from one another
- “in stock” shows all product that have stock 
- “available” shows all the products that can be added to cart
- “not available” shows all out-of-stock products  
-> tested on all types of products, with different quantities in combination, tested in different categories/subcategories
- changes Availability preferences in product page X times, refreshed page, Availability option is correctly updated
- filtered from faceted search left column, active filter horizontal block and URL
- tested on multiple types of filter (Availability, Property, Brand, Price etc)
- tested with option Allow ordering of out-of-stock products enabled and disabled

For me it is a go ! Thanks a lot for your work !

Progi1984:
Thanks @Hlavtox & @florine2623

rdy4ever:
Hi @Hlavtox , @Progi1984 
This is a PR I've been waiting for some time and thanks for making it possible. However, it looks like you missed the case when the product quantity is < 0 and backorder is enabled. In this case, the products should still be "Available", right?

Hlavtox:
@rdy4ever Yes, this should work.
Is something not working? 🤔

Hlavtox:
Ohh, I see what you mean. I will test tommorow. ;)

rdy4ever:
> Ohh, I see what you mean. I will test tommorow. ;)

Yeap, 
When a product has backorders active, its stock gets negative, not 0. And if backorders are allowed, it should still be "Available".
Oh, and about the wording, wouldn't "Backorder" make more sense than "Available" for this filter? "Available" might be a bit missleading for some folks and still think it's in stock.

rdy4ever:
> Oh, and about the wording, wouldn't "Backorder" make more sense than "Available" for this filter? "Available" might be a bit missleading for some folks and still think it's in stock.

Nevermind this part. I see both in stock and backorder are included in "Available"

Hlavtox:
Well, Available will get you all products that should be orderable. If we changed the wording to “backorder”, we would need to exclude in stock products.

Regarding the thing, deleting this line will solve it if my thoughts are right - `['quantity', [0], '>='],`

hibatallahAouadni:
Indeed it works!

rdy4ever:
> edit: it works

It works, but you might need to do sometging about the count in the () too, as it still shows only > 0 results
image
I have 14 results and only 5 shown inside the filter

rdy4ever:
Oh, and you should apply that change for the multiple selection too (when both "in stock" and "available" are selected)

Hlavtox:
Yes :) Counter and two cases in selection

PR #531, opened 2021-10-18
Ability for module to work on all pages

Hlavtox:
| Questions     | Answers
| Description?  | This PR allows the module to work on all controllers.
| Type?         | improvement
| BC breaks?    | ?
| Deprecations? | no
| Fixed ticket? | Fixes PrestaShop/Prestashop#15137
| How to test?  | Install module and see that filter works on all pages.

## Functionality
- In every template, user will select controllers where he wants to use this filter template. Category checkboxes are needed only if using the template on category page.
- In the filter overview, merchant can nicely see which pages use the filter template he created.
- On special pages
  - If using category filter, it always sets the home category as the root.
  - Settings **Show products from subcategories** and **Show products only from default category are ignored** 

## Details
- I added a controller column to filter table.
- Supports all default product listing controllers. 
- Supports any custom controller specified in the list. We can make some hook for this possibly. ;-)
- Technically, the biggest change is `WHERE id_category = " . $idCategory` to `WHERE controller = 'category' AND id_category = " . $idCategory` when getting filters. The other stuff was mainly to adjust admin and some conditions here and there.
- I separated the rendering of admin screen to separate functions at least. It would be good to move everything related to Admin and Indexing to separate pages, but that is not the scope of this PR.
- Cache
  - If the page should be cached or not is determined by 'cacheable' property in controller list.
  - We are currently caching category, manufacturer, supplier and search pages. We could omit search, because it could grow big.
  - We can't cache new product page because we don't know if the product expired from the page. The same with best-sales and price-drop.
  - Module does not need https://github.com/PrestaShop/PrestaShop/pull/26285 to work, I added backward compatibility function which can be later removed.

## Screenshots
fo
bo1
bo2

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/ps_facetedsearch/531)

### DIFF
--- a/ps_facetedsearch.php
+++ b/ps_facetedsearch.php
@@ -72,7 +72,7 @@ class Ps_Facetedsearch extends Module implements WidgetInterface
      *
      * @var array
      */
-    private $available_controllers = [];
+    public $available_controllers = [];
 
     /**
      * @var bool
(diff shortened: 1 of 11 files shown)


matks:
Hi @Hlavtox maybe you opened too many topics 😉  this one did not move for 30 days so we're supposed to close it. Can I close it?

PierreRambaud:
Hi @Hlavtox ,

Since we had no news from you for more than 30 days, I'll close this pull request. Feel free to reopen or open another one if you think it's still relevant.

Thanks!

PR #569, opened 2021-12-12
Filtering basd on average Review/Comment scores and Sorting 

samberrry:
| Questions     | Answers
| Description?  | Adds the ability to filter products based on average comment scores and sorting. This feature works with native PrestaShop comments module
| Type?         | new feature
| BC breaks?    | yes / no
| Deprecations? | no
| Fixed ticket? | 
| How to test?  |


samberrry:
Hi guys, 
I worked on this feature couple months ago, and I thought that it would be a good idea to be added to faceted search module as it's a nice feature for shops, so I decided to make a PR for that :)
thank you

samberrry:
Screen Shot 2021-12-12 at 7 13 29 PM

samberrry:
I will add more explanation to the Description or in comments ;) thanks

ps_facetedsearch.php
```diff
@@ -165,6 +165,10 @@ protected function getDefaultFilters()
                 'label' => 'Product price filter (slider)',
                 'slider' => true,
             ],
+            'layered_selection_review_star' => [
```
PierreRambaud, line 168: This option should be available only when productcomments is installed.

ps_facetedsearch.php
```diff
@@ -1162,6 +1170,8 @@ public function buildLayeredCategories()
                             } elseif (substr($key, 0, 23) == 'layered_selection_feat_') {
                                 $sqlInsert .= '(' . (int) $idCategory . ', ' . (int) $idShop . ', ' . (int) str_replace('layered_selection_feat_', '', $key) . ',
 \'id_feature\',' . (int) $n . ', ' . (int) $limit . ', ' . (int) $type . '),';
+                            } elseif ($key == 'layered_selection_review_star') {
+                                $sqlInsert .= '(' . (int) $idCategory . ', ' . (int) $idShop . ', NULL,\'review\',' . (int) $n . ', ' . (int) $limit . ', ' . (int) $type . '),';
```
PierreRambaud, line 1174:
```suggestion
                                $sqlInsert .= '(' . (int) $idCategory . ', ' . (int) $idShop . ', NULL, \'review\', ' . (int) $n . ', ' . (int) $limit . ', ' . (int) $type . '),';
```

ps_facetedsearch.php
```diff
@@ -1503,4 +1513,88 @@ public function getWidgetVariables($hookName, array $configuration)
...
+    {
+        $this->getDatabase()->execute('DROP TABLE IF EXISTS `' . _DB_PREFIX_ . 'layered_comment_index`');
+
+        $this->getDatabase()->execute('DROP TABLE IF EXISTS `' . _DB_PREFIX_ . 'layered_comment_index_log`');
+
+        $this->getDatabase()->execute(
+            'CREATE TABLE `' . _DB_PREFIX_ . 'layered_comment_index` (
+            `id_product` INT  NOT NULL,
+            `score` INT NOT NULL,
+            `avg_score` FLOAT NOT NULL,
```
PierreRambaud, line 1530: Do we really need to use Float like that maybe we can do 
```suggestion
            `avg_score` FLOAT(5, 4) NOT NULL,
```
Meaning we can do 9.9999
No need to do 25.390843984 ^^

ps_facetedsearch.php
```diff
@@ -1503,4 +1513,88 @@ public function getWidgetVariables($hookName, array $configuration)
...
+     */
+    public function rebuildCommentIndexTable()
+    {
+        $this->getDatabase()->execute('DROP TABLE IF EXISTS `' . _DB_PREFIX_ . 'layered_comment_index`');
+
+        $this->getDatabase()->execute('DROP TABLE IF EXISTS `' . _DB_PREFIX_ . 'layered_comment_index_log`');
+
+        $this->getDatabase()->execute(
+            'CREATE TABLE `' . _DB_PREFIX_ . 'layered_comment_index` (
+            `id_product` INT  NOT NULL,
```
PierreRambaud, line 1528:
```suggestion
            `id_product` INT NOT NULL,
```

ps_facetedsearch.php
```diff
@@ -1503,4 +1513,88 @@ public function getWidgetVariables($hookName, array $configuration)
...
+            PRIMARY KEY (`id_product`),
+            INDEX `score` (`score`),
+            INDEX `avg_score` (`score`)
+            ) ENGINE=' . _MYSQL_ENGINE_ . ' DEFAULT CHARSET=utf8;'
+        );
+
+        $this->getDatabase()->execute(
+            'CREATE TABLE `' . _DB_PREFIX_ . 'layered_comment_index_log` (
+            `id_comment` INT  NOT NULL,
+            `id_product` INT  NOT NULL,
```
PierreRambaud, line 1540:
```suggestion
            `id_comment` INT NOT NULL,
            `id_product` INT NOT NULL,
```

ps_facetedsearch.php
```diff
@@ -1503,4 +1513,88 @@ public function getWidgetVariables($hookName, array $configuration)
...
+
+            foreach ($this->getDatabase()->executeS($query) as $commentLog) {
+                $this->addCommentIndexLog($commentLog);
+            }
+
+        } else {
+            $query = 'select pc.id_product_comment, pc.id_product, pc.grade from ps_product_comment as pc
+left join ps_layered_comment_index_log as lc
+on pc.id_product_comment = lc.id_comment
+where lc.indexed is null;';
```
PierreRambaud, line 1571:
```suggestion
            $query = 'SELECT pc.id_product_comment, pc.id_product, pc.grade ' .
                'FROM ' . _DB_PREFIX_ . '_product_comment as pc ' .
                'LEFT JOIN ' . _DB_PREFIX_ . '_layered_comment_index_log as lc ' .
                'pc.id_product_comment = lc.id_comment ' .
                'WHERE lc.indexed IS NULL';
```

Easier to read and fix error with unknown table begining with `ps_`  prefix

src/Filters/Block.php
```diff
@@ -964,4 +967,69 @@ private function preparePriceSpecifications()
...
+        $query = 'SELECT  count(g.grade) as count ,g.grade FROM (SELECT t.id_product, case 
+when t.avg_grade between 1 and 1.99 then "1"
+when t.avg_grade between 2 and 2.99 then "2"
+when t.avg_grade between 3 and 3.99 then "3"
+when t.avg_grade between 4 and 5 then "4"
+end as grade
+ FROM (SELECT avg(h.grade) as avg_grade,
+ h.id_product FROM 
+ (select pc.id_product, pc.grade from ' . _DB_PREFIX_ . 'product_comment as pc
+inner join (SELECT * FROM ' . _DB_PREFIX_ . 'category_product where id_category = '.$_GET['id_category'].') cp
```
PierreRambaud, line 997: Sql injection here, you should use `$idParent` defined in line 97

src/Filters/Block.php
```diff
@@ -964,4 +967,69 @@ private function preparePriceSpecifications()
...
+    {
+        $values = [
+            '4' => ['name'=>"4"],
+            '3' => ['name'=>"3"],
+            '2' =>  ['name'=>"2"],
+            '1' => ['name'=>"1"],
+        ];
+
+        $query = 'SELECT  count(g.grade) as count ,g.grade FROM (SELECT t.id_product, case 
+when t.avg_grade between 1 and 1.99 then "1"
```
PierreRambaud, line 989: We shouldn't be able to do 0.5 star? I think it costs nothing to start at 0 instead of 1

src/Filters/Block.php
```diff
@@ -964,4 +967,69 @@ private function preparePriceSpecifications()
...
+when t.avg_grade between 2 and 2.99 then "2"
+when t.avg_grade between 3 and 3.99 then "3"
+when t.avg_grade between 4 and 5 then "4"
+end as grade
+ FROM (SELECT avg(h.grade) as avg_grade,
```
PierreRambaud, line 994: Two much subqueries, I'm afraid of performance problems :thinking:

src/Hook/ProductComment.php
```diff
@@ -0,0 +1,71 @@
...
+        $grade = $params['object'];
+
+        $commentLogRow = $this->database->executeS('SELECT * FROM `' . _DB_PREFIX_ . 'layered_comment_index_log`
+            WHERE id_comment = ' . $grade->id . ' AND indexed = 1');
+
+        if (empty($commentLogRow)){
+            $gradeCommentRow = $this->database->executeS('SELECT * FROM `' . _DB_PREFIX_ . 'layered_comment_index`
+            WHERE id_product =' . $grade->id_product);
+
+            if (empty($gradeCommentRow) || !$gradeCommentRow){
```
PierreRambaud, line 47:
```suggestion
            if (empty($gradeCommentRow)){
```

src/Product/Search.php
```diff
@@ -233,6 +233,18 @@ private function addSearchFilters($selectedFilters, $parent, $idShop)
...
                     }
                     break;
+
+                case 'review':
+                    if (isset($selectedFilters['review'])
+                        && (
+                            $selectedFilters['review'][0] !== ''
+                        )
+                    ) {
+                        $this->addReviewFilter(
```
PierreRambaud, line 243: You should be able to put more than one review

PierreRambaud (changes requested):
Hey thank you @samberrry that's a good idea, but I think there are missing important things before being able to merge it.
You can't select two stars, which is something we should be able to do like "I want all 4 and 5 stars items"
image
Seen SQL request are buggy ^^
image

I'll continue the review after ;)

samberrry:
Hi @PierreRambaud, Thanks for review and your comments. 
Sorry I did not have a chance to explain in details, so there are some misunderstandings. 
Before doing a comment/review filter it's important to figure out what is our measurement? actually how we want to filter and prune product result..
we have multiple options to choose: filter products based on number of comments, based on sum of scores or average scores, etc.. and you know they have different indexing.. 

so, I've implemented something like what Amazon does in filtering:
https://www.amazon.com/s?k=Home+bedding&i=kitchen-intl-ship&bbn=16225011011&rh=p_72%3A1248917011&dc&pd_rd_r=6a42ef25-9ece-45e9-956f-aafdc97e2421&pd_rd_w=FAkBn&pd_rd_wg=YaJWn&pf_rd_p=bbbef69f-3d9e-41ab-b1eb-e08d5690939b&pf_rd_r=27756W73C6EYFG572GVS&qid=1639410469&rnid=1248913011&ref=sr_nr_p_72_3
Screen Shot 2021-12-13 at 10 08 57 AM

we do not filter based on counts, so when we click on this:
Screen Shot 2021-12-13 at 10 25 26 AM
we expect to see products which have average score equal to "2" or higher stars. This 'two' shows that there are two available products with '2' stars. It's an 'up' filtering, this filter is not multiple choices (like others), we do not expect to have both '**products with average 4 stars and up**' and '**products with average 1 star and up'**. 

Also the number 3 in here:
Screen Shot 2021-12-13 at 10 33 51 AM
 It shows that the '3 stars Up' filter is selected, it's just an identifier, I should check that if it's possible change it to '**Avg. Customer Average Reviews: 3 Up**' 

Again thanks for your review, I should fix the items you mentioned on code.. actually I could not work on the feature too much to optimized and review it very well, we still have some considerations in some sections which is one of the important issues, and it's about indexing. It works well when a moderator must approve comments, because I do indexing on comment validated action hook, we also need to do indexing on comment addition when the moderator checking is disabled. 
I will continue to work on it. 

Thanks ;)

ps_facetedsearch.php
samberrry: yes, will add.

ps_facetedsearch.php
samberrry: correct :D

ps_facetedsearch.php
samberrry: correct thanks.

src/Filters/Block.php
samberrry: will add.

src/Filters/Block.php
samberrry: actually the average score can not be less than 1, because we do not have comments with 0 grade, user must choose between 1-5 values.

src/Product/Search.php
samberrry: explained in comments bellow.

src/Filters/Block.php
samberrry: done

src/Filters/Block.php
samberrry: yes you are right, I think at least 3 joins are inevitable, but an alternative could be another table which keeps counts for all categories, and can be calculated during indexing.

PierreRambaud:
I really like your idea and better understand the reason about the ` & Up` :sweat_smile: 
I'll try to help you as much as I can :+1:

samberrry:
@PierreRambaud Thank you 😊
I try to make it better too 😉

ps_facetedsearch.php
samberrry: done

samberrry:
Hey @PierreRambaud, I was just working on comment indexing and I noticed that the product comment module is using doctrine ORM to store the product comment models, and there is not action hook to get comment addition event or delete event. We need it for the case that the moderator approval is disabled:
Screen Shot 2021-12-15 at 5 07 03 PM
In this situation any comment will be shown without approval, and we need to index these comments somehow. We have action hook for comment approval, so we can update index very well, but when the approval is disabled we do not have events from comment addition as the model is handled by doctrine.
I think the only way is to set cron job, I don know if there is another way to get these events.
I would be happy to know your opinion 😉 thanks

PierreRambaud:
> Hey @PierreRambaud, I was just working on comment indexing and I noticed that the product comment module is using doctrine ORM to store the product comment models, and there is not action hook to get comment addition event or delete event. We need it for the case that the moderator approval is disabled: Screen Shot 2021-12-15 at 5 07 03 PM In this situation any comment will be shown without approval, and we need to index these comments somehow. We have action hook for comment approval, so we can update index very well, but when the approval is disabled we do not have events from comment addition as the model is handled by doctrine. I think the only way is to set cron job, I don know if there is another way to get these events. I would be happy to know your opinion wink thanks

We have two possible choices:
- Use only SQL queries to retrieve the status of each comment
- Reduce compatibility range and use Doctrine in this one too

I'm not sure we should remove the compatibility range of this module since it's use by a lot of people, maybe playing with SQL should be enough

samberrry:
> > Hey @PierreRambaud, I was just working on comment indexing and I noticed that the product comment module is using doctrine ORM to store the product comment models, and there is not action hook to get comment addition event or delete event. We need it for the case that the moderator approval is disabled: Screen Shot 2021-12-15 at 5 07 03 PM In this situation any comment will be shown without approval, and we need to index these comments somehow. We have action hook for comment approval, so we can update index very well, but when the approval is disabled we do not have events from comment addition as the model is handled by doctrine. I think the only way is to set cron job, I don know if there is another way to get these events. I would be happy to know your opinion wink thanks
> 
> We have two possible choices:
> 
>     * Use only SQL queries to retrieve the status of each comment
> 
>     * Reduce compatibility range and use Doctrine in this one too
> 
> 
> I'm not sure we should remove the compatibility range of this module since it's use by a lot of people, maybe playing with SQL should be enough

yes. But how should we get notified on comment addition/deletion even?
now there is not any event to trigger indexing.. I think we have another option to do that. We can add an action hook in product comment module, so we can get the event and index it for filtering. 
So, I think we have two options, one is cron job, and the second is action hook within product comments module.
If it's necessary, I will try to make a PR on product comments module to add this action hook.

Hlavtox:
@samberrry @PierreRambaud Guys, this will be a NICE addition! Impressive 🚀

Few points from myself:
- Try to maintain performance, ideally combine it in one single query as it is now. It's already not very optimized, we must keep performance a priority. :-)
- What about if you hide other ratings, if you select one? The filter would reappear if the user deletes the selected filter from "Active filters".

kpodemski:
Hey @samberrry

It's a great contribution. I know you're probably busy, but will you continue with it? It's been a while since the last activity there.

samberrry:
> Hey @samberrry
> 
> It's a great contribution. I know you're probably busy, but will you continue with it? It's been a while since the last activity there.

Hey @kpodemski,

of course. I should change the indexing process so that it support the situation when the admin approval is disabled. Now it depends on admin approval hook. I realized that I'm blocked with product comments Doctrine (as I've said in previous comments), but now I've just checked product comment module and I see that it's ObjectModel and I can get addition and deletion events, seems that I did not notice. However I may need your help in some parts. Let me check all aspects on next week.

Thanks ❤️

samberrry:
I checked it, and as I see in productcomments module, entity manger is being used to store Product Comments objects. So, the actionObjectProductCommentAddAfter will be never called. That's what I was talking about. 

I will make a PR on productcomments and will add two hooks for adding and removing comments. These are necessary hooks to compute indexes on adding and removing comments events.

samberrry:
> Hello @samberrry
> 
> Sorry for the delay , Pierre is no longer at PrestaShop he was following this PR.
> 
> I see that you target the master branch can you change? And solve the conflicts.

Hello @okom3pom 
no worries. Sure, I will do. It seems that I should update it with latest master changes.

samberrry:
- git rebase (to the latest master version) and conflict resolution done. 
- I've not tested yet.
- After doing tests I will mention reviewers.

samberrry:
> @samberrry PRs must be submitted against `dev` branch, not `master` 😉 I changed the base branch of this Pull Request

Oh thanks! mistaken :)

PR #593, opened 2022-01-15
Fix `strip_tags` error messages during value escaping.

FongWan:
| Questions     | Answers
| Description?  | This PR fixes the issue when an array-type value is given to the `pSQL` function, since the `$values` variable could be a multi-dimensional array.
| Type?         | bug fix
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes PrestaShop/Prestashop#27357.
| How to test?  | Please see PrestaShop/Prestashop#27357.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/ps_facetedsearch/593)

### DIFF
--- a/src/Adapter/MySQL.php
+++ b/src/Adapter/MySQL.php
@@ -316,11 +316,17 @@ class MySQL extends AbstractAdapter
      */
     protected function getJoinedEscapedValue($separator, array $values)
     {
-        return implode($separator, array_map(function ($value) use ($separator) {
-            return is_array($value) ?
-                $this->getJoinedEscapedValue($separator, $value) :
-                (is_numeric($value) ? pSQL($value) : "'" . pSQL($value) . "'");
-        }, $values));
+        foreach ($values as $key => $value) {
+            if (is_array($value)) {
+                $values[$key] = $this->getJoinedEscapedValue($separator, $value);
+            } elseif (is_numeric($value)) {
+                $values[$key] = pSQL($value);
+            } else {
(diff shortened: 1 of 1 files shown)


src/Adapter/MySQL.php
```diff
@@ -306,6 +306,23 @@ protected function getFieldMapping()
...
+     * Get the joined and escaped value from an multi-dimensional array
+     *
+     * @param string $separator
+     * @param array $values
+     *
+     * @return string Escaped string value
+     */
+    protected function getJoinedEscapedValue($separator, array $values)
+    {
+        return implode($separator, array_map(function ($value) use ($separator) {
```
PierreRambaud, line 319: Using anonymous function is not the best for performance :thinking:

FongWan: @PierreRambaud Although I wanted to reply you with "Premature optimization is the root of all evil" 🤣, but you are right, anonymous functions in this case can increase up to two times the time required.

So, I have optimized the function to use `foreach` instead of using `array_map` with a callback function. This could reduce up to 40% of the time required.

PierreRambaud (approved):

HanaRebaiQA (approved):
Hello @FongWan 

I have tested this PR. The issue is fixed.
I checked on PS1782, PS178x and develop branches + multistore.
https://watch.screencastify.com/v/pnbl1cZHM1cJ1C2SPhvH

So, QA :heavy_check_mark: 

Thanks!

PR #718, opened 2022-09-28
Fixed Price And URL When Filtering With Combinations

TheKruspe:
- Added `id_product_attribute` To Select-Fields

Edit: I just noticed the product default combination gets ignored if a filter based on attribute groups is active.

| Questions     | Answers
| Description?  | When filtering products with combinations, only the default combination was used to determine price and url of the results.
| Type?         | bug fix / improvement
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes PrestaShop/Prestashop#15738 and PrestaShop/Prestashop#15164
| How to test?  | Filter for products with multiple combinations and different prices. Click on a product to navigate to the correct combination.

- - -
This change is [Reviewable](https://reviewable.io/reviews/prestashop/ps_facetedsearch/718)

### DIFF
--- a/ps_facetedsearch.php
+++ b/ps_facetedsearch.php
@@ -198,6 +198,7 @@ class Ps_Facetedsearch extends Module implements WidgetInterface
             Configuration::updateValue('PS_LAYERED_FILTER_PRICE_ROUNDING', 1);
             Configuration::updateValue('PS_LAYERED_FILTER_SHOW_OUT_OF_STOCK_LAST', 0);
             Configuration::updateValue('PS_LAYERED_FILTER_BY_DEFAULT_CATEGORY', 0);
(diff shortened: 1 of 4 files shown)


matks (changes requested):
Hi @TheKruspe the CI is red can you check why tests are failing?

TheKruspe:
I just made changes that restore the default combination behaviour when using attribute filters (mentioned in edit). 

Furthermore, it is now possible to display every combination of a product as a separate product. I'm not sure if I should leave this feature in this branch. I will now check why tests are failing.

TheKruspe:
Some tests need to be updated, because I modified `useFiltersAsInitialPopulation` . I'll fix and update later this week.

kpodemski:
@Hlavtox have you seen this PR? shall we take over it?

TheKruspe:
@kpodemski I unfortunately wasn't able to find the time. If you have any questions, let me know.

Hlavtox:
@TheKruspe Implemented in develop branch. :-)

PR #726, opened 2022-10-26
Remove use of `$_GET`, refacto, simplify

Hlavtox:
| Questions     | Answers
| Description?  | See below
| Type?         | improvement / refacto
| BC breaks?    | yes
| Deprecations? | no
| Fixed ticket? | 
| How to test?  | This Pull Request refactors the code, but does not modify the behavior. You should be able to compare the module behavior with and without this PR : they should be identical

### Intro for reviewers on how this works
- When you open a category or other page, the controller first asks on a hook, if there is somebody who would like to search products for him. This module is hooked there and if it's category controller, it says "Hey I can do this filtering for you" and return it's instance.
- Then, category controller calls `runQuery()` function in this module and passes `ProductSearchQuery` object. This object contains all the information about category ID, selected filters, query type, page, sorting etc. 
- This basically means = "Hey `ps_facetedsearch`, you agreed to search products for me, so get me the products then. Use these instructions."

It looks like this:
test

### What was wrong
- Even though all the required information is provided in the `ProductSearchQuery`, ps_facetedsearch was using it only partially and in some places in the code, it was doing `Tools::getValue('id_category');` which is not very reliable.
- This is a problem when you want to fetch those products from a different custom controller, you then need to do weird stuff like 
`$_POST['id_category'] = 471;`
- So I changed this and now it takes the category ID from `ProductSearchQuery` properly everywhere.
- It also make implementing https://github.com/PrestaShop/ps_facetedsearch/pull/531 much easier, because I will have all the query data available everywhere properly.

### What I changed 
- src/Filters/Block - Added `$query` to constructor and using `$this->query->getIdCategory()`.
- src/Filters/Converter - Query already provided, so I just used `$query->getIdCategory()``.
- src/Filters/Products - Instead of passing the query data in 4 variables (`$productsPerPage`, `$page`, `$orderBy`, `$orderWay`) I passed only `$query` like in `src/Filters/Block`.
- src/Hook/ProductSearch - Just some comments and early return to avoid big `if` block.
- src/Product/Search - Added `$query` property, getter, setter. So I can use `$this->query->getIdCategory();`. Initialize the Category object only if psLayeredFullTree is true, no need to do it in the beginning if it's not used eventually.
- src/Product/SearchProvider - Mostly comments, adding setQuery call and changing two calls.

- - -
This change is [Reviewable](https://reviewable.io/reviews/prestashop/ps_facetedsearch/726)

### DIFF
--- a/src/Hook/ProductSearch.php
+++ b/src/Hook/ProductSearch.php
@@ -33,7 +33,7 @@ class ProductSearch extends AbstractHook
     ];
 
     /**
-     * This function returns the search provider to the controller who requested it.
+     * This method returns the search provider to the controller who requested it.
      * Module currently only supports filtering in categories, so in other cases,
      * we don't return anything.
      *
--- a/src/Hook/ProductSearch.php
+++ b/src/Hook/ProductSearch.php
@@ -33,7 +33,9 @@ class ProductSearch extends AbstractHook
     ];
 
     /**
-     * Hook project search provider
+     * This function returns the search provider to the controller who requested it.
+     * Module currently only supports filtering in categories, so in other cases,
+     * we don't return anything.
      *
      * @param array $params
      *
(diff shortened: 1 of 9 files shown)


src/Hook/ProductSearch.php
```diff
@@ -33,50 +33,48 @@ class ProductSearch extends AbstractHook
     ];
 
     /**
-     * Hook project search provider
+     * This function returns the search provider to the controller who requested it.
```
marsaldev, line 36:
```suggestion
     * This method returns the search provider to the controller who requested it.
```

marsaldev (dismissed):

marsaldev (approved):

mflasquin (approved):
Hi @Hlavtox, it seems good :) 
Maybe you should add more informations about the page / feature to test in the "How to test" section for the QA ?
Thanks :smile:

HanaRebaiQA:
Hello @Hlavtox 

The CI tests are not successful. We should start testing when everything is OK
image

Hlavtox:
Hi @HanaRebaiQA, the failing test is irrelevant. The coveralls test checks code coverage by tests, which decreased, even though I actually fixed some tests with this PR. 😄 @kpodemski pls confirm 🙏

HanaRebaiQA:
Hello @Hlavtox 
That's OK for us.

MhiriFaten (approved):
Hello @Hlavtox ,

I have checked your PR and it is QA approved :heavy_check_mark: 

Thank you !

Hlavtox:
@MhiriFaten @HanaRebaiQA Thank you! 🚀

PR #727, opened 2022-10-27
Cache features in DataAccessor

Hlavtox:
| Questions     | Answers
| Description?  | Read below
| Type?         | refacto
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | 
| How to test?  | Compare installs with and without this PR, they will provide the same results.

### Description
- There is no need to load a list of features or their values 30 times.
- There was already a mechanism to store loaded attributes, I implemented the same for features.

### Stats
- I tried on my shop with a category that has 2592 products, 28 features and about 800 feature values.
- Average load time from 8 attempts 507 ms ➡️ 492 ms
- Database queries 386 ➡️ 359
- Peak memory usage 15,4 MB ➡️ 15,3 MB

### How to test
- I think this can be tested by a dev. There is no difference to the returned data.

- - -
This change is [Reviewable](https://reviewable.io/reviews/prestashop/ps_facetedsearch/727)

### DIFF
--- a/src/Filters/DataAccessor.php
+++ b/src/Filters/DataAccessor.php
@@ -32,12 +32,22 @@ class DataAccessor
     /**
      * @var array
      */
-    private $attributesGroup;
+    private $attributesGroup = [];
 
     /**
      * @var array
      */
-    private $attributes;
+    private $attributes = [];
+
+    /**
+     * @var array
+     */
+    private $features = [];
+
+    /**
+     * @var array
+     */
+    private $featureValues = [];
 
     /**
      * @var Db
@@ -50,9 +60,11 @@ class DataAccessor
     }
 
     /**
+     * Get all attributes for a given language and attribute group.
+     *
      * @param int $idLang
      *
-     * @return array|false|\PDOStatement|resource|null
+     * @return array Attributes
      */
     public function getAttributes($idLang, $idAttributeGroup)
     {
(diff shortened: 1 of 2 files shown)


src/Filters/DataAccessor.php
```diff
@@ -39,6 +39,16 @@ class DataAccessor
      */
     private $attributes;
 
+    /**
+     * @var array
+     */
+    private $features;
```
zuk3975, line 45:
```suggestion
    private $features = [];
```

src/Filters/DataAccessor.php
```diff
@@ -39,6 +39,16 @@ class DataAccessor
...
+
+    /**
+     * @var array
+     */
+    private $featureValues;
```
zuk3975, line 50:
```suggestion
    private $featureValues = [];
```
just to avoid possibility of multiple types? (cuz by default its null)

marsaldev (approved):

zuk3975 (approved):

zuk3975:
smth seems to gone wrong with the phpstan though

Hlavtox:
@zuk3975 It fails randomly on every run, see other PRs, the same situation. 🤷

kpodemski:
@PrestaShop/prestashop-core-developers 

Any Docker magician over there? this is the error I noticed some time ago. It fails randomly on directory deletion

MhiriFaten (commented):
Hello @Hlavtox ,

I tested the PR and the installations with and without this PR give the same results.
So it's OK :heavy_check_mark: 

:warning:  But some checks were not successful ! 
image

Ping @PrestaShop/prestashop-maintainers could you check it please? 

Thank you !

Hlavtox:
@MhiriFaten @jolelievre Tests green we can merge! ✌️

zuk3975 (approved):

PR #728, opened 2022-10-27
Add default filter functionality

Hlavtox:
| Questions     | Answers
| Description?  | Read below. I tried to provide clear description of the feature with context and how it looks like. Please ask me if some things are not clear.
| Type?         | new feature
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes https://github.com/PrestaShop/PrestaShop/issues/28687
| How to test?  | Read below. I tried to provide clear 'how to test' items please ask me if some things are not clear.

### Description
- Half of merchants have no idea about any filter settings and you must non-stop keep checking, if you didn't forget to assign a template to new category after creating it. Merchants don't want to think about it.
- So, this PR adds a new function where you can select a default filter that you want to assign to new categories. After creating a new category, if this function is enabled, the category will be added to this filter.
- It also refactors the admin a tiny bit, adds some comments, extracts some logic to separate functions and fixes a tiny bug.

### How does it look
template

### How to test and what to test
Firstly, this PR modifies a behavior that happens when adding a new category. Then some parts of the code were extracted to separate functions to make it SOMEHOW cleaner. It concerns only the backend part, there cannot be any effect on filtering in front office.

**So, to test default filter functionality:**
- Set up a filter in faceted search for some categories. 
- Set up another filter for another categories.
- Select one of the filters as your default one.
- Then go to Catalog > Categories and create a new category.
- Go back to faceted search and see that this category should be automatically added to your default filter. The other filter should stay exactly as it was.
- Check that nothing else was changed in that filter other than the 1 new category.

**A tiny bug fixed**
- If you delete a category, faceted search goes through each filter template and deletes the category from the filter template. But, it forgots to update the category count in the filter. Although maybe used only for BO, it's good if everything is correct.
- Setup a filter template for some categories. Then go to Catalog > Categories and delete one of the categories.
- See that on dev branch, category count in filter template list stayed the same. With my PR, it gets correctly updated.

**General test**
- Click through all the settings in the module settings, see that everything works.
- Try adding a new filter template, try editing some template, try deleting some template.
- Check that you are offered all your features and attributes for filtering.

You can briefly check FO, but there is no effect.

- - -
This change is [Reviewable](https://reviewable.io/reviews/prestashop/ps_facetedsearch/728)

### DIFF
--- a/ps_facetedsearch.php
+++ b/ps_facetedsearch.php
@@ -803,13 +803,14 @@ class Ps_Facetedsearch extends Module implements WidgetInterface
         $features = $this->getAvailableFeatures();
         $attributeGroups = $this->getAvailableAttributes();
 
-        // Initialize category tree
+        // Initialize category tree component
         $treeCategoriesHelper = new HelperTreeCategories('categories-treeview');
         $treeCategoriesHelper
             ->setRootCategory((Shop::getContext() == Shop::CONTEXT_SHOP ? Category::getRootCategory()->id_category : 0))
             ->setUseCheckBox(true);
 
-        // If we are editing an already existing template
+        // If we are editing an already existing template, we will load its data,
+        // check categories and add selected filters. Otherwise, we prepare empty template.
         if ($template !== null) {
             $filters = Tools::unSerialize($template['filters']);
             $id_layered_filter = $template['id_layered_filter'];
@@ -837,7 +838,7 @@ class Ps_Facetedsearch extends Module implements WidgetInterface
         }
 
         $this->context->smarty->assign([
-            'current_url' => $this->context->link->getAdminLink('AdminModules') . '&configure=ps_facetedsearch&tab_module=front_office_features&module_name=ps_facetedsearch',
+            'current_url' => $this->context->link->getAdminLink('AdminModules', true, [], ['configure' => $this->name, 'tab_module' => $this->tab, 'module_name' => $this->name]),
             'id_layered_filter' => $id_layered_filter,
             'template_name' => $template_name,
             'attribute_groups' => $attributeGroups,
@@ -879,7 +880,7 @@ class Ps_Facetedsearch extends Module implements WidgetInterface
     }
 
     /**
-     * Returns array with all available attributes on the shop
+     * Returns array with all available attributes on the shop. Only used in backoffice.
      */
     private function getAvailableAttributes()
     {
@@ -894,7 +895,7 @@ class Ps_Facetedsearch extends Module implements WidgetInterface
     }
 
     /**
-     * Returns array with all available features on the shop
+     * Returns array with all available features on the shop. Only used in backoffice.
      */
     private function getAvailableFeatures()
     {
@@ -1580,7 +1581,7 @@ VALUES(' . $last_id . ', ' . (int) $idShop . ')');
     }
 
     /**
-     * Provides data about single template.
+     * Provides data about single filter template.
      *
      * @param int $idFilterTemplate ID of filter template
      *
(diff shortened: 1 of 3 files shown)


ps_facetedsearch.php
```diff
@@ -1105,43 +1160,57 @@ public function rebuildLayeredCache($productsIds = [], $categoriesIds = [], $reb
     }
 
     /**
-     * Build layered categories
+     * This function gets serialized data of filter templates from layered_filter table and builds detailed
```
marsaldev, line 1163:
```suggestion
     * This method gets serialized data of filter templates from layered_filter table and builds detailed
```

marsaldev (approved):

ps_facetedsearch.php
```diff
@@ -834,11 +788,74 @@ public function getContent()
...
+        if (Shop::isFeatureActive() && count(Shop::getShops(true, null, true)) > 1) {
+            $helper = new HelperForm();
+            $helper->id = Tools::getValue('id_layered_filter', null);
+            $helper->table = 'layered_filter';
+            $helper->identifier = 'id_layered_filter';
+            $this->context->smarty->assign('asso_shops', $helper->renderAssoShop());
+        }
+
+        $this->context->smarty->assign([
+            'current_url' => $this->context->link->getAdminLink('AdminModules') . '&configure=ps_facetedsearch&tab_module=front_office_features&module_name=ps_facetedsearch',
```
matks, line 840: Suggestion: use https://www.php.net/manual/en/function.http-build-query.php

ps_facetedsearch.php
```diff
@@ -1506,4 +1578,52 @@ public function getWidgetVariables($hookName, array $configuration)
...
+     * @param int $idFilterTemplate ID of filter template
+     *
+     * @return array Filter data
+     */
+    public function getFilterTemplate($idFilterTemplate)
+    {
+        return $this->getDatabase()->getRow(
+            'SELECT *
+            FROM `' . _DB_PREFIX_ . 'layered_filter`
+            WHERE id_layered_filter = ' . $idFilterTemplate
```
matks, line 1594:
```suggestion
            WHERE id_layered_filter = ' . (int) $idFilterTemplate
```
Against SQL injection

matks (commented):
@Hlavtox This PR has stayed here for a while so I think it's pretty clear very few people feel confortable enough to say "Yes, I believe this PR is OK and it introduces no regressions".

This ps_facetedsearch module is quite complex and it's easy to introduce a bug when modifying the code, because every piece of code is linked to the others. It is mostly due to complex data processing and cache optimization.

But I don't think we should block this contribution because we are unable to verify it fully. If we do that, we never go forward, right? More and more I believe that instead of halting contributions that have a risk of introducing a problem, we should be confident in the wonderful QA team to find it and also in our ability to release a bugfix if we find a problem later.

So the only change I request is the little integer casting (for SQL injection) then I'll approve the PR.

Hlavtox:
Hi @matks! :-) Thank you for feedback, yes, it's dangerous grounds, but I dived deep into it recently and I think I have at least some idea how it works now. 😄

I took your feedback into account and:
- Fixed that SQL injection danger
- Generated the URL by internal mechanisms instead of manually
- Found and fixed a tiny bug when deleting a shop category

matks (approved):
Let's go 👍

sallemiines (commented):
Hello @Hlavtox 
Thanx for the PR ! it's LGTM QA :heavy_check_mark: 

- Tests auto : OK 
You can see the attached screen record 

[video-13451289-10847f1e90cab513a65a3df46115eaf8.webm](https://user-images.githubusercontent.com/31663369/208885369-137c67e5-f193-4f8a-a0d8-7d15a66da2ea.webm)

Thank you !

Hlavtox:
Yaaaaay, thanks @sallemiines & myself

PR #754, opened 2022-12-15
Fix missing products if default order is by manufacturer

Hlavtox:
| Questions     | Answers
| Description?  | If you set default order in category by manufacturer and the product has no manufacturer, its not visible at all. It is because the manufacturer table has an INNER JOIN. If you LEFT JOIN it, everything works correctly. Doesn't affect filtering.
| Type?         | bug fix
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes PrestaShop/Prestashop#30562
| How to test?  | See below or see the issue.

**How to test**
- Prepare a category, where there will be some product from manufacturer A, some products from manufacturer B and some without it.
- Go to Shop Parameters > Product Settings and set default sorting by Brand.
- Go check that category and see that the products without manufacturer disappeared.
- With my PR, everything works correctly.
- Only thing that this change could break is manufacturer filtering, checked, works correctly.

- - -
This change is [Reviewable](https://reviewable.io/reviews/prestashop/ps_facetedsearch/754)


lartist (approved):

nicosomb (approved):

MhiriFaten (approved):
Hello @Hlavtox ,

I have checked your PR !
The issue is well fixed and everything works correctly :heavy_check_mark: 

Thank you.

kpodemski:
thanks @Hlavtox :)

PR #757, opened 2022-12-21
Ability for module to work on more pages

Hlavtox:
qa

| Questions     | Answers
| Description?  | See below
| Type?         | new feature
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes PrestaShop/Prestashop#15137
| How to test?  | See below

# What this is about
- Customer can now filter on all pages! On manufacturer, supplier, prices drop, best sales, new products, search page.
- You now have extra options in module settings, where you assign which type of pages you want this filter template to be used on.
- Category checkboxes are now needed only if using the template on category page.
- In the filter overview, merchant can nicely see which pages use the filter template he created.
- There is a new column in the database generated filters that marks which controller it belongs to.

# How to test
- Test with classic theme.
- Absolutely no change was done on category pages. Just check that the filters show there and it shows some products. 
- Check if the new controllers show proper products, if sorting works and there are no new bugs specific to this new functionality. Check it on classic theme, there are some issues on hummingbird.
- There are some common issues in this module that we know about, when you find something, always compare to develop or ask me on Slack:
  - https://github.com/PrestaShop/PrestaShop/issues/29727 - test only with level of subcategories 1 in module settings.
  - https://github.com/PrestaShop/PrestaShop/issues/26803
  - https://github.com/PrestaShop/PrestaShop/issues/12501

# How does it look
filter templates
template

# Changes made
### STEP 1 - adding controllers, administration, serialized filter templates and reacting properly:
- [x] Add supported controllers to module.
- [x] Add inputs and overview to admin.
- [x] Process admin data, extend serialized filter data by controller information.
- [x] Provide search interface only in case of supported controller.

### STEP 2 - Add controller column, generate filters, make proper SQL queries:
- [x] Modify "generated filters" DB structure, generate it with new data, get this data with a controller condition. Add upgrade.
- [x] Take new controllers into account when generating filters.

### STEP 3 - Use the new data when filtering:
- [x] Setup proper caching depending on the page.
- [x] Use correct filter data on the new pages. 
- [x] Add new conditions for new controllers that have joins.
- [x] Add new conditions for new controllers that don't have joins.
- [x] Check and implement sorting options on new controllers.

### STEP 4 - Checks and polishing:
- [x] Add compatibility with 1.7.6-1.7.8 and their missing queryTypes.
- [x] Verify product selection behavior on special pages.
- [x] Add tests for Search class.

### Issues on Hummingbird (already being fixed on theme side)
- Using price filter on search page breaks it, it removes `s` parameter from the URL.
- Removing filters on new products page for example removes them, but it keeps the old URL.

### Issues found on the core side
- On best sales page, the default sorting without this module is `sales, desc`, but with this module it's `name, asc`. 
- It's because in BestSalesControllerCore, there is `'product', 'name'` set on the query, which arrives to our module. 
- But `BestSalesProductSearchProvider` completely ignores it and overrides it by `'product', 'sales'`, which we don't have a clue about.
- **I fixed this in the module, so you won't notice, but should be fixed also on core side (https://github.com/PrestaShop/PrestaShop/pull/31074).**

- - -
This change is [Reviewable](https://reviewable.io/reviews/PrestaShop/ps_facetedsearch/757)


marsaldev (commented):
I've tested this first step, LGTM

ps_facetedsearch.php
```diff
@@ -1595,4 +1651,67 @@ public function getFilterTemplate($idFilterTemplate)
...
+
+    /**
+     * Should this controller filter blocks be cached?
+     */
+    public function shouldCacheController($controller)
+    {
+        return $this->supportedControllers[$controller]['cacheable'];
+    }
+
+    public function initializeSupportedControllers()
```
kpodemski, line 1684: It probably doesn't need to be an initializer. Why don't just have `getSupportedControllers(): array`? Later on, in a different PR, we could think about hooks to make it flexible.

ps_facetedsearch.php
```diff
@@ -1595,4 +1651,67 @@ public function getFilterTemplate($idFilterTemplate)
...
+    {
+        return isset($this->supportedControllers[$controller]);
+    }
+
+    /**
+     * Should this controller filter blocks be cached?
+     */
+    public function shouldCacheController($controller)
+    {
+        return $this->supportedControllers[$controller]['cacheable'];
```
kpodemski, line 1681:
```suggestion
        return $this->supportedControllers[$controller]['cacheable'] ?? false;
```

just to make sure

ps_facetedsearch.php
```diff
@@ -1595,4 +1651,67 @@ public function getFilterTemplate($idFilterTemplate)
...
+     */
+    public function isControllerSupported($controller)
+    {
+        return isset($this->supportedControllers[$controller]);
+    }
+
+    /**
+     * Should this controller filter blocks be cached?
+     */
+    public function shouldCacheController($controller)
```
kpodemski, line 1679:
```suggestion
    public function shouldCacheController(string $controller): bool
```

src/Filters/Provider.php
```diff
@@ -56,7 +56,8 @@ public function getFiltersForQuery(ProductSearchQuery $query, int $idShop)
         if (empty($this->filters)) {
             $this->filters = $this->database->executeS(
             'SELECT type, id_value, filter_show_limit, filter_type FROM ' . _DB_PREFIX_ . 'layered_category
-            WHERE id_category = ' . (int) $query->getIdCategory() . '
+            WHERE controller = \'' . $query->getQueryType() . '\'
```
kpodemski, line 59:
```suggestion
            WHERE controller = "' . pSQL($query->getQueryType()) . '"
```

src/Product/Search.php
```diff
@@ -132,6 +134,30 @@ public function initSearch($selectedFilters)
...
     }
 
+    public function getProductIdsUsingCoreSearch()
+    {
+        // Search using the core functionality
+        $result = SearchCore::find(
+            $this->context->language->id,
+            Tools::replaceAccentedChars(urldecode($this->query->getSearchString())),
+            1,
+            2147483647, // Limit is the 32bit integer
```
kpodemski, line 144:
```suggestion
            PHP_INT_MAX,
```

src/Product/SearchProvider.php
```diff
@@ -83,24 +83,28 @@ public function __construct(
     }
 
     /**
+     * @param ProductSearchQuery $query
+     *
      * @return array
      */
-    private function getAvailableSortOrders()
+    private function getAvailableSortOrders($query)
```
kpodemski, line 90: Not blocking, but if querytype is only what we need, we could ask for query type.

src/Product/SearchProvider.php
```diff
@@ -213,13 +232,25 @@ public function runQuery(
...
+
+        $filterKey = $query->getQueryType();
+        if ($query->getQueryType() == 'category') {
+            $filterKey .= $query->getIdCategory();
+        } elseif ($query->getQueryType() == 'manufacturer') {
+            $filterKey .= $query->getIdManufacturer();
+        } elseif ($query->getQueryType() == 'supplier') {
+            $filterKey .= $query->getIdSupplier();
+        } elseif ($query->getQueryType() == 'search') {
+            $filterKey .= $query->getSearchString() . $query->getSearchTag();
```
kpodemski, line 244: Search is not cachable anyway, right? I don't know exactly how it works, but generating a cache for every single search string would be overkill.

views/templates/admin/manage.tpl
```diff
@@ -66,6 +67,7 @@
 			<tr>
 			  <td class="center">{(int)$template['id_layered_filter']}</td>
 			  <td class="text-left">{$template['name']}</td>
+				<td>{$template['controllers']}</td>
```
kpodemski, line 70:
```suggestion
			  <td class="center">{(int)$template['id_layered_filter']}</td>
			  <td class="text-left">{$template['name']}</td>
			  <td>{$template['controllers']}</td>
```

src/Product/SearchProvider.php
Hlavtox: Fixed

src/Filters/Provider.php
Hlavtox: `Single quotes should be used for string values like in the VALUES() list. Double quotes are supported by MySQL for string values as well, but single quotes are more widely accepted by other RDBMS, so it is a good habit to use single quotes instead of double.`

src/Product/SearchProvider.php
Hlavtox: I will pass the whole query, maybe we want to do other things with it?

kpodemski (dismissed):

MhiriFaten (commented):
Hello,

We can not test this PR because there are some checks were not successful.
I will remove the 'waiting for QA' label.

Thanks!

kpodemski:
@MhiriFaten read PR table, **please!**

kpodemski (approved):

src/Product/Search.php
```diff
@@ -132,6 +134,30 @@ public function initSearch($selectedFilters)
...
+            1,
+            PHP_INT_MAX,
+            'position',
+            'desc',
+            false,
+            false,
+            null
+        );
+
+        // Extract IDs, return some nonsense
```
zuk3975, line 152: explain more in this comment, or remove it.
Now it just raises questions why are you returning "nonsense"

kpodemski (approved):

hibatallahAouadni (approved):
Hello @Hlavtox 

The block is added and we can manage where the filter block is displayed:

image

It looks like a charm:

[pr_757_b.webm](https://user-images.githubusercontent.com/23308427/216373773-4ca26f07-b00c-4457-9e29-85b0663d326b.webm)

Except the responsive display which is outscoop of this PR.
So, LGTM and QA :heavy_check_mark: 

Thanks!

Hlavtox:
AAAAAAYYYYYYYYYYYYYYYYYYYY

kpodemski:
🎉 🎉 🎉 🎉 🎉 🎉

PR #775, opened 2023-01-18
Refacto applying filters, add tests

Hlavtox:
➡️ QA - read the PR description and test scenarios provided.

| Questions     | Answers
| Description?  | Read below.
| Type?         | refacto
| BC breaks?    | no
| Deprecations? | 
| Fixed ticket? | 
| How to test?  | Try test scenarios below.

### What it's about
- This is another extracted part of https://github.com/PrestaShop/ps_facetedsearch/pull/757. 
- Basically, we have three "kinds" of filters that are applied to a search.
  - Ones that are common for all pages (even for other controllers potentially) and are applied all the time, no matter what.
  - Then you have filters that customers select in front office by themselves in left column.
  - And you have the filters specific for the current controller, currently only a category page.
- It's a mess now and it needs to be cleaned up a bit. So I got to work:

### What you should know to approve and understand the code
- When "Show full tree is enabled" in settings, the module uses `nleft`/`nright` filters. If not, it uses specific `id_category` filter.
- If "Show only products from subcategories" is enabled, it also adds `id_category_default` filter.

### What I did
- User specified filters stay in `addSearchFilters`. That's good.
- I moved all common filters to `addCommonFilters`. This concerns shop ID, group limit and product visibility.
- Page specific filters are moved to a new function called `addControllerSpecificFilters`.
  - The logic is slightly changed. Before it was like taking three right turns instead of one left. :-)
  - Before, it was applying `id_category_default` filter first and then removing it again by calling `resetFilter()` Now, it checks if any user selected category filter was applied, and only if not, it sets the `id_category_default` filter and other filters like `nleft` and `nright`.
- Fixed a test error that appeared. When I submit a search with user selected filters containing category `6`, why there should be categories `6` and `12` in expected answer? There should be `id_category=6` filter only.
- Added a new test that checks if proper `nleft`/`nright` filters are added instead of id_category, depending on `PS_LAYERED_FULL_TREE` configuration.

### Why I do this in this separate PR
- In https://github.com/PrestaShop/ps_facetedsearch/pull/757, I will just modify one `addControllerSpecificFilters` function and it will be much easier to review and troubleshoot.
- It's much easier to troubleshoot and add tests when other controllers are not mixing in.

### What to test in QA
- **You don't have to check backoffice, nothing was done there.**
- Scenario 1
  - Put one product only to category `Women`. **Don't put it in Clothes**.
  - Enable `Show products from subcategories`, go to category `Clothes`, see that you can see this product.
  - Disable `Show products from subcategories`, go to category `Clothes`, see that you cannot see this product.
- Scenario 2
  - Put one product to categories `Men` and `Women`. Set default category of this product to `Women`.
  - Enable `Show products only from default category`, see that you can see this product only in `Women` and not in `Men`.
  - Disable `Show products only from default category`, see that you can see this product both in `Women` and `Men` categories.

### Github is not very good at showing the difference, it looks much cleaner IRL 👍
code

- - -
This change is [Reviewable](https://reviewable.io/reviews/PrestaShop/ps_facetedsearch/775)

### DIFF
--- a/src/Product/Search.php
+++ b/src/Product/Search.php
@@ -118,47 +118,27 @@ class Search
      */
     public function initSearch($selectedFilters)
     {
-        // Get category ID from the query or home category as a fallback
-        $idCategory = (int) $this->query->getIdCategory();
-        if (empty($idCategory)) {
-            $idCategory = (int) Configuration::get('PS_HOME_CATEGORY');
-        }
+        // Adds basic filters that are common for every search, like shop and group limitations
+        $this->addCommonFilters();
 
-        $psLayeredFullTree = Configuration::get('PS_LAYERED_FULL_TREE');
-        if (!$psLayeredFullTree) {
-            $this->addFilter('id_category', [$idCategory]);
-        }
+        // Add filters that the user has selected for current query
+        $this->addSearchFilters($selectedFilters);
 
-        $psLayeredFilterByDefaultCategory = Configuration::get('PS_LAYERED_FILTER_BY_DEFAULT_CATEGORY');
-        if ($psLayeredFilterByDefaultCategory) {
-            $this->addFilter('id_category_default', [$idCategory]);
-        }
+        // Adds filters that specific for category page
+        $this->addControllerSpecificFilters();
 
-        // Visibility of a product must be in catalog or both (search & catalog)
-        $this->addFilter('visibility', ['both', 'catalog']);
-
-        // User must belong to one of the groups that can access the product
-        if (Group::isFeatureActive()) {
-            $groups = FrontController::getCurrentCustomerGroups();
-
-            $this->addFilter('id_group', empty($groups) ? [Group::getCurrent()->id] : $groups);
-        }
-
-        $this->addSearchFilters(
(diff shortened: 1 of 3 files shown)


Hlavtox:
➡️ QA - read the PR description and test scenarios provided.

marsaldev (approved):

zuk3975:
nice detailed description 👏 💪

MhiriFaten (approved):
Hello @Hlavtox ,

I have checked your PR and it looks OK !

**Scenario 1**

[Untitled_ Jan 19, 2023 4_28 PM.webm](https://user-images.githubusercontent.com/111756615/213482675-52aa71c8-4703-446b-b487-a7948c290436.webm)

**Scenario 2**

[Untitled_ Jan 19, 2023 4_31 PM.webm](https://user-images.githubusercontent.com/111756615/213483432-074f06b1-d571-4521-adff-e9dd80b7fbb3.webm)

So it's QA approved :heavy_check_mark: 

Many thanks!

Hlavtox:
All perfect then! Thank you!

PR #789, opened 2023-02-09
Make search as fast as original provider

Hlavtox:
| Questions     | Answers
| Description?  | Instead of leveraging the search controller, which is not optimal for our usecase, I reimplemented a part of core search functionality to the module itself, which will make it as fast as the original provider.
| Type?         |  improvement
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | 
| How to test?  | See below.

**How to test**
- _You don't need to test other pages than search, they were not touched._
- _You don't need to test the filtering in the left column._
- Order of the products could be different compared to when the module is disabled, because the sorting by relevance is not yet implemented.
- **Do test** that the search page finds the same results as if the module is disabled. Try it on 1.7.6 and 8.0.1.

**Example of a performance boost**
Before
- Load Time 3457 ms - You'd better run your shop on a toaster
- Querying Time 2628 ms
- Queries 14950

After
- Load Time 660 ms - Unicorn powered webserver!
- Querying Time 457 ms
- Queries 1505

- - -
This change is [Reviewable](https://reviewable.io/reviews/PrestaShop/ps_facetedsearch/789)

### DIFF
--- a/src/Product/CoreSearchBackport.php
+++ b/src/Product/CoreSearchBackport.php
@@ -1,0 +1,413 @@
+<?php
+
+/**
+ * Copyright since 2007 PrestaShop SA and Contributors
+ * PrestaShop is an International Registered Trademark & Property of PrestaShop SA
+ *
+ * NOTICE OF LICENSE
+ *
+ * This source file is subject to the Academic Free License 3.0 (AFL-3.0)
+ * that is bundled with this package in the file LICENSE.md.
+ * It is also available through the world-wide-web at this URL:
+ * https://opensource.org/licenses/AFL-3.0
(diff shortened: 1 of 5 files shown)


kpodemski (approved):
As discussed, it's a workaround, but it's better than nothing. For v9, we can make sure that SearchController will share the found products IDs 😎

kpodemski (approved):

sallemiines:
Hello @Hlavtox 

Thanx for the PR , it's LGTM QA :heavy_check_mark: 

**search for a product ( shop without PR )**

image

**search for a product ( shop with PR )**

image

Thank you :rocket:

PR #779, opened 2023-01-31
Use numeric IDs for filter values (concept)

Hlavtox:
🚧 Work in progress - will remove also the url_name forms, attribute indexing etc.

| Questions     | Answers
| ------------- | -------------------------------------------------------
| Description?  | Uses unique values for filter values - feature values, attribute values, categories and others. This fixes many things - category filter for deeper categories, values with the same name, fixes all problems with URL encoding.
| Type?         | refacto
| BC breaks?    | yes - behavior changed
| Deprecations? | -
| Fixed ticket? | Fixes PrestaShop/Prestashop#22637, Fixes PrestaShop/Prestashop#23726, Fixes PrestaShop/Prestashop#15635, Fixes PrestaShop/Prestashop#29727
| How to test?  | Play around with module and see that filtering still works fine.

Snímek obrazovky 2023-01-31 163011

- - -
This change is [Reviewable](https://reviewable.io/reviews/PrestaShop/ps_facetedsearch/779)

PR #793, opened 2023-02-14
Fix Converter.php for non category pages

PrestaConnect:
Fixes the bug that prevented filtering by category from working except for category pages. There is no need for parent category ID control, except for category pages.

| Questions     | Answers
| Description?  | In line with the newly added feature of the ps_facetedsearch module, it has also been enabled to work on Bestsellers, Price Drops, New products and Search pages. However, while the category filter was running on these pages, except for the category pages, the category filter was not working because it was checked whether the category was a sub-category of the current level category.
| Type?         | bug fix
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | No follow up issue as it is outside of PrestaShop project
| How to test?  | On the new products page, you can test by trying filtering for a category that is not child to the main category.

- - -
This change is [Reviewable](https://reviewable.io/reviews/PrestaShop/ps_facetedsearch/793)

### DIFF
--- a/src/Filters/Converter.php
+++ b/src/Filters/Converter.php
@@ -400,7 +400,11 @@ class Converter
                 case self::TYPE_CATEGORY:
                     if (isset($facetAndFiltersLabels[$filterLabel])) {
                         foreach ($facetAndFiltersLabels[$filterLabel] as $queryFilter) {
-                            $categories = Category::searchByNameAndParentCategoryId($idLang, $queryFilter, (int) $idCategory);
+                            if ($query->getQueryType() == 'category') {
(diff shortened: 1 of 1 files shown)


Hlavtox:
@PrestaConnect Hi, on other pages, the home category is 2, so it works.
The category is filter is buggy because it selects the category by it's name.

If you have Clothes > Big and Accessories > Big, if you filter by Big, it won't work.

The only proper solution is to add IDs into URL, I am preparing it here - https://github.com/PrestaShop/ps_facetedsearch/pull/779 👍

PrestaConnect:
Thank you for your work.
I am closing my pull request.
I wish a good day.

PR #857, opened 2023-06-20
Hook to allow modify faceted seach query

borjagarciab:
| Questions     | Answers
| Description?  | I've seen that currently is not possible modify the query from faceted search module. This will allow external modules to modify the query to achieve its functionality.
| Type?         |  improvement / new feature
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | https://github.com/PrestaShop/PrestaShop/issues/32951
| How to test?  | Execute this hook and modify the array you need. 
||image

- - -
This change is [Reviewable](https://reviewable.io/reviews/PrestaShop/ps_facetedsearch/857)

### DIFF
--- a/src/Adapter/MySQL.php
+++ b/src/Adapter/MySQL.php
@@ -327,6 +327,8 @@ class MySQL extends AbstractAdapter
                 'joinType' => self::INNER_JOIN,
             ],
         ];
+        
+        \Hook::exec('actionFSQueryBuilderModifier', ['filterToTableMapping' => &$filterToTableMapping]);
 
         return $filterToTableMapping;
     }


Hlavtox (commented):
Hi @borjagarciab, this only allows to modify the table mapping, not the query. We should have more thought about it. :-)

borjagarciab:
Hi @Hlavtox 
Maybe it's not the best hook name, but by modifying these tables, we can allow to modify the conditions without a major update....
By modifying the array "joinCondition" we can add an additional where for example.
It is possible to modify the getQuery function to use the DbQuery class (for example), but that is a major update....

Currently i can't see other way to modify this query...
Please check it.
Thanks!

jocel1:
For me the MySQL adapter is not the right place to add filtering / changes to the queries. It's designed to be as generic as possible and provide basic functionalities to allow other adapter to be created, like for example a ElasticSearch adapter. 
If you want to add new filtering, you should do this at a higher level.

kpodemski:
Since we have had no news from you for more than 20 days, I'm closing this PR. Feel free to re-open it if you find the motivation to finish it. Don't hesitate to ping me if you have any questions or doubts.

PR #873, opened 2023-07-07
Add filters checkbox state in editing template;

fatm-dev:
| Questions     | Answers
| Description?  | Added check if filter's id is in filter array in template editing filter checkboxes. Moved layered_selection_ag to new variable cause its repeating 3 times.
| Type?         | bug fix
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes PrestaShop/Prestashop#32987
| How to test?  | See PrestaShop/Prestashop#32987


Hlavtox:
Will be fixed by https://github.com/PrestaShop/ps_facetedsearch/pull/881 - the checkboxes are checked by javascript.

PR #880, opened 2023-07-17
Add possibility to filter by discounts, on sale and new products

Hlavtox:
| Questions     | Answers
| Description?  | Adds a new feature that allows filtering for new products, discounted products or products with "On sale" flag.
| Type?         | new feature
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | 
| How to test?  | Try this new filter, try to upgrade the module.

### Description
- Adds a new feature that adds possibility to filter by some new product special things.
  - You can filter out products that are new.
  - You can filter out products that are discounted.
  - You can filter out products that have "On sale" label on them.
- I named a group of these new features "Highlights".

### Screenshot
Snímek obrazovky 2023-07-17 094147

- - -
This change is [Reviewable](https://reviewable.io/reviews/PrestaShop/ps_facetedsearch/880)


ps_facetedsearch.php
```diff
@@ -1191,6 +1194,12 @@ public function rebuildLayeredCache($productsIds = [], $categoriesIds = [], $reb
                     $doneCategories[(int) $idCategory]['p'] = true;
                     $toInsert = true;
                 }
+
+                if (!isset($doneCategories[(int) $idCategory]['q'])) {
+                    $filterData['layered_selection_highlights'] = ['filter_type' => Converter::WIDGET_TYPE_CHECKBOX, 'filter_show_limit' => 0];
+                    $doneCategories[(int) $idCategory]['e'] = true;
```
kpodemski, line 1200: `e` like "extra properties"? 🤔

kpodemski (commented):
Only one question.

I kinda like "Highlights", but a few alternatives I see:
- Products types (although it's confusing with product types like pack, combination, etc.)
- Products selections
- Shopping choices
- Specific products

ps_facetedsearch.php
Hlavtox: Yep, forgotten

MatShir (changes requested):
Small request: We should separate the new and the discounts part 🙏

kpodemski:
@MatShir what do you mean? it is about showing products with different state, on discount, on sale, new products - how would you like to split this?

l-delin:
Hello @Hlavtox :-)

As @MatShir said, it would be nice to have two separate parts for the two "types" of labels so it's not too confusing. By that, I mean that `Discounted` & `On sale` products should **not** be in the same section as `New products`, because they don't share common properties. 
I liked your proposition with the use of "highlights" but in an ecommerce context, it doesn't work. We need to use words users are familiar with. 
There's a second point: how can we translate it? It works fine in English, but for languages like French, you'd need to split it into two words and it would take too much space.

So, considering this, here's what we propose:

Instead of the `Highlights` category, two categories: `Deals` and `Selection`. 

The **Deals** section is dedicated to `Discounted` and `On sale` products. 
And, as a personal suggestion: Ultimately, "Flash sales" could be added to this category as well.

The **Selection** section contains a selection of specific products. In this case, the `New products` category. 
Again, suggestion: categories such as season-themed or holiday-themed product collections could be added to this category. (e.g.: "Christmas collection", "Summer collection", etc.)

So it would look like this:

##
**Deals:**
 - Discounted
 - On Sale

##
**Selection:**
 - New products


What do you think? 🏎

src/Adapter/MySQL.php
```diff
@@ -93,23 +93,26 @@ public function execute()
      */
     public function getQuery()
     {
+        // Prepare mapping for joined tables
         $filterToTableMapping = $this->getFieldMapping();
+
+        // Process and generate all fields for the SQL query below
         $orderField = $this->computeOrderByField($filterToTableMapping);
+        $selectFields = $this->computeSelectFields($filterToTableMapping);
```
Hlavtox, line 101: This is just a cosmetic change that has no effect, but the computations should be together.

src/Adapter/MySQL.php
```diff
@@ -218,6 +221,7 @@ protected function getFieldMapping()
                 'fieldName' => 'name',
                 'joinCondition' => '(p.id_manufacturer = m.id_manufacturer)',
                 'joinType' => self::LEFT_JOIN,
+                'dependencyField' => 'id_manufacturer',
```
Hlavtox, line 224: Safety precatution, I want to add the select dynamically, and if id_manufacturer is not in initial population, this mapping would not work, because the adapter doesn't know what `(p.id_manufacturer = m.id_manufacturer)` is.

src/Adapter/MySQL.php
```diff
@@ -324,7 +328,7 @@ protected function getFieldMapping()
                     (sp.from = \'0000-00-00 00:00:00\' OR \'' . date('Y-m-d H:i:s') . '\' >= sp.from) AND 
                     (sp.to = \'0000-00-00 00:00:00\' OR \'' . date('Y-m-d H:i:s') . '\' <= sp.to) 
                 )',
-                'joinType' => self::INNER_JOIN,
+                'joinType' => self::LEFT_JOIN,
```
Hlavtox, line 331: INNER_JOIN was a mistake, we are checking for a value in WHERE. We can't inner join if the table is in an OR clause.

src/Product/SearchProvider.php
```diff
@@ -542,16 +537,15 @@ private function hideUselessFacets(array $facets, $totalProducts)
...
-                 */
+                // There is only one filter, but it reduces the product selection
                 (
                     count($facet->getFilters()) === 1
                     && $totalFacetProducts < $totalProducts
                     && $usefulFiltersCount > 0
                 )
+                ||
+                // There is only one filter, but it's availability or highlights
+                ($usefulFiltersCount === 1 && ($facet->getType() == 'availability' || $facet->getType() == 'highlights'))
```
Hlavtox, line 548: This is a bug, we must check if there is at least one useful filter, or it will display empty blocks on a page when there no products.

Hlavtox:
@MatShir @kpodemski It's ready for review.
- I went for facet name "Selections" and putting three values there. After Slack discussion, it doesn't make sense to have a separate category with one filter.
- This filter is now enabled on new module install. For existing users, it's not.
- When installing this module, I moved the stock, price and selections filter to the top. As they are the most important.
- The filters work like this, if enabled - ((`On sale` OR `Discounted`) AND `New product`)

**How it looks**
Snímek obrazovky 2023-09-27 142901

MatShir:
Looks good to me !

MatShir (approved):

boherm (approved):

Hlavtox:
Hi @florine2623, it works only on clean installs when there never was faceted search. For existing installs, there will be no change.

florine2623 (approved):
@Hlavtox ,

Indeed after Uninstalling the module and reinstalling with the PR? the 3 elements are on top the the filter list ^^

QA ✅ 
Thanks!

kpodemski:
well done 👏🏻 thanks @Hlavtox

MatShir:
> Hi @florine2623, it works only on clean installs when there never was faceted search. For existing installs, there will be no change.

Even after a module upgrade ? what needs to be done for this part to be on top ? @Hlavtox

PR #883, opened 2023-07-18
Improve availability filter

Hlavtox:
| Questions     | Answers
| Description?  | Several small improvemets to availability filters. Details below.
| Type?         | improvement
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Doesn't fix an issue, but will fix about thousand complains from merchants.
| How to test?  | Check that stock filter works correctly.

### Changes
- Many people complain that they have three values - "Not available", "Available" and "In stock". Now, if the products Available and In stock are the same, it will don't display "Available".
- I removed automatic sorting by language for availability. Now they will be fixed in this order - "In stock", "Available", "Not available".
- Even if "In stock" is the last option with products, it will still be displayed, so people know if the X products remaining are in stock or not. Not everybody has stock displayed in their product list.
- Changed the keys of those values underground to textual ones. 0,1,2 doesn't really do anything and it's not understandable.


kpodemski (approved):

src/Filters/Block.php
```diff
@@ -509,6 +509,13 @@ private function getAvailabilitiesBlock($filter, $selectedFilters)
                     }
                 }
             }
+
+            // Hide Available option if the count is the same as In stock, it doesn't make no sense
+            // Product count is a reliable indicator here, because there can never be product IN STOCK that is not AVAILABLE
+            // So if the counts match, it MUST BE the same products
+            if ($availabilityOptions[1]['nbr'] == $availabilityOptions[2]['nbr']) {
+                unset($availabilityOptions[1]);
+            }
```
matthieu-rolland, line 518: I understand the idea here...  but I find it weird that the filter would display the same elements depending on whether "available" and "in stock" elements have the same list

it kinda breaks the UI consistency to have sometimes both and sometimes "available" disappears...  

but maybe I'm wrong, I'm not a UX expert

has this been validated with the product team?

matthieu-rolland (commented):
just some uneducated feedback...  and also I'm wondering if this should be validated by @MatShir before being QAed and merged

AureRita:
Hi @Hlavtox 

Thank you for your PR, All seems to works as you can see here : 

[Untitled_ Jul 24, 2023 5_03 PM.webm](https://github.com/PrestaShop/ps_facetedsearch/assets/100699445/535081e2-6e00-4cfe-980f-6331f131fe7b)

Tested on :
1.7.8.10 :heavy_check_mark: 
8.1.1 :heavy_check_mark: 
8.0.5 :heavy_check_mark: 

I see just a little issue on it :
If you've already load the previous faceted search, when you update, if you don't the state of any product, the faceted search'll continue to be like before, only in your languages, as you can see : 

[Untitled_ Jul 24, 2023 5_31 PM.webm](https://github.com/PrestaShop/ps_facetedsearch/assets/100699445/fe7a354d-cb92-4535-95d1-0b64c314ccbf)

I don't think this one'll block this PR so it's QA :heavy_check_mark:

MatShir:
🥳

PR #945, opened 2023-11-11
Fix date formatting

Hlavtox:
| Questions     | Answers
| Description?  | We were happily calling a function with a wrong argument count. This fixes it.
| Type?         | bug fix
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes tests in https://github.com/PrestaShop/PrestaShop/pull/34521, which revealed this issue when I typed displayDate method.
| How to test?  | Dev QA.


matthieu-rolland (approved):

Hlavtox:
@florine2623 Added some steps to reproduce. :-)

florine2623 (commented):
Hello @Hlavtox ,

I can't see the new date format 😞 

I have followed your steps. Also, cleared cache from PS and from browser as well. The old format is still displayed.
Screenshot 2023-11-14 at 15 10 06

Am I missing something ? ^^

Hlavtox:
Yes, just the fact that you don't have any errors makes this QA ✅. :-)

florine2623:
Yes indeed, the date format changes on your country. I didn't think about that, sorry ^^

PR #977, opened 2024-02-01
Update SearchProvider.php

Prestaworks:
| Questions     | Answers
| Description?  | Change translation domain for making it possible to change values in BO with module translation.
| Type?         | bug fix
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes PrestaShop/Prestashop#35237.
| How to test?  | Follow instructions in Issue report

### DIFF
--- a/src/Product/SearchProvider.php
+++ b/src/Product/SearchProvider.php
@@ -103,37 +103,37 @@ class SearchProvider implements FacetsRendererInterface, ProductSearchProviderIn
 
         $sortOrders = [
             $sortSalesDesc->setLabel(
-                $translator->trans('Sales, highest to lowest', [], 'Shop.Theme.Catalog')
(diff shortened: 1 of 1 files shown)


Hlavtox (commented):
Why you want to do this? We want to retain the same translations as in the core, no?
This will only result in having to translate these strings again.

Prestaworks:
Two of those strings does not exist in core so it's impossible to translate them. Maybe because Core translations does not pickup from modules folder? But generally it's better to keep translations for modules with the module. If not, it's just hard to find where the translation is supposed to happen.

Hlavtox (changes requested):
https://github.com/PrestaShop/PrestaShop/issues/35237#issuecomment-1923599109

PR #983, opened 2024-02-13
Deprecate standalone endpoints to match PrestaShop 9.0 new security policy

jolelievre:
| Questions     | Answers
| Description?  | Allow executing only the specific PHP files as external endpoints, required since PrestaShop 9.0 because additional security was added in this PR https://github.com/PrestaShop/PrestaShop/pull/34184 The idea in this PR is to allow access to very specific files to prevent unwanted accesses on malicious files
| Type?         | bug fix
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | ~
| How to test?  | Install a shop based on develop branch, go to the module administration of faceted search and try to build indexes or clear the cache. An error message should appear and when you check the network in browser developer tools you should see that the ajax request was forbidden (403 error). Now use the branch from this PR and do the same, the ajax calls should now be successful.

### DIFF
--- a/controllers/front/cron.php
+++ b/controllers/front/cron.php
@@ -1,0 +1,80 @@
+<?php
+/**
+ * Copyright since 2007 PrestaShop SA and Contributors
+ * PrestaShop is an International Registered Trademark & Property of PrestaShop SA
+ *
+ * NOTICE OF LICENSE
+ *
+ * This source file is subject to the Open Software License (OSL 3.0)
+ * that is bundled with this package in the file LICENSE.md.
+ * It is also available through the world-wide-web at this URL:
+ * https://opensource.org/licenses/OSL-3.0
+ * If you did not receive a copy of the license and are unable to
+ * obtain it through the world-wide-web, please send an email
+ * to license@prestashop.com so we can send you a copy immediately.
+ *
+ * DISCLAIMER
+ *
+ * Do not edit or add to this file if you wish to upgrade PrestaShop to newer
+ * versions in the future. If you wish to customize PrestaShop for your
+ * needs please refer to https://devdocs.prestashop.com/ for more information.
+ *
+ * @author    PrestaShop SA and Contributors <contact@prestashop.com>
+ * @copyright Since 2007 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ */
+
+declare(strict_types=1);
+
+class Ps_FacetedSearchCronModuleFrontController extends ModuleFrontController
+{
+    public function __construct()
+    {
+        parent::__construct();
+        $this->ajax = true;
+    }
+
+    public function postProcess()
+    {
+        if (substr(Tools::hash('ps_facetedsearch/index'), 0, 10) != Tools::getValue('token')) {
+            header('HTTP/1.1 403 Forbidden');
+            header('Status: 403 Forbidden');
+            $this->ajaxRender('Bad token');
+
+            return;
+        }
+
+        $action = Tools::getValue('action');
+        switch ($action) {
+            case 'indexAttributes':
+                Shop::setContext(Shop::CONTEXT_ALL);
+
+                $psFacetedsearch = new Ps_Facetedsearch();
+                $psFacetedsearch->indexAttributes();
+                $psFacetedsearch->indexFeatures();
+                $psFacetedsearch->indexAttributeGroup();
+
+                $this->ajaxRender('1');
+                break;
+            case 'clearCache':
+                $psFacetedsearch = new Ps_Facetedsearch();
+                $this->ajaxRender($psFacetedsearch->invalidateLayeredFilterBlockCache());
+                break;
+            case 'indexPrices':
+                Shop::setContext(Shop::CONTEXT_ALL);
+
+                $module = new Ps_Facetedsearch();
+                if (Tools::getValue('full')) {
+                    $this->ajaxRender($module->fullPricesIndexProcess((int) Tools::getValue('cursor'), (bool) Tools::getValue('ajax'), true));
+                } else {
+                    $this->ajaxRender($module->pricesIndexProcess((int) Tools::getValue('cursor'), (bool) Tools::getValue('ajax')));
+                }
+
+                break;
+            default:
+                header('HTTP/1.1 403 Forbidden');
+                header('Status: 403 Forbidden');
+                $this->ajaxRender('Unknown action');
+        }
+    }
+}
--- a/ps_facetedsearch-attribute-indexer.php
+++ b/ps_facetedsearch-attribute-indexer.php
@@ -17,6 +17,14 @@
  * @copyright Since 2007 PrestaShop SA and Contributors
  * @license   https://opensource.org/licenses/AFL-3.0 Academic Free License 3.0 (AFL-3.0)
  */
+
+/*
+ * This standalone endpoint is deprecated, it should not be used anymore and should be removed along with the
+ * htaccess file that still allows it to work despite the security policy from the core forbidding this kind
+ * of file to be executed.
+ */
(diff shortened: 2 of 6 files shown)


kpodemski:
My understanding is that we should create ModuleFrontControllers as a replacement for .PHP files, not to show how to omit the rules for developers in the ecosystem :D

kpodemski:
I'll also repeat what @matks already wrote:

> I hope the people who participated in this discussion to advocate for this change will be helping this task to be complete 😝 . I plan to submit a few PRs to fix some of the modules listed but I will not do all of them.

maybe we should ping @jf-viguier @clotairer @touchweb-vincent

boherm (approved):

Hlavtox:
@kpodemski With French efficiency, we will be in grave by then :D

Hlavtox:
@jolelievre Inspiration here - https://github.com/PrestaShop/gsitemap/pull/175/files

Just throw some deprecation messages and create three simple files like:

```
class Ps_facetedsearchPriceindexerModuleFrontController extends ModuleFrontController
{
    public function postProcess()
    {
      if (substr(Tools::hash('ps_facetedsearch/index'), 0, 10) != Tools::getValue('token')) {
          exit('Bad token');
      }

      Shop::setContext(Shop::CONTEXT_ALL);
      $this->module->indexAttributes();
      $this->module->indexFeatures();
      $this->module->indexAttributeGroup();
    }
}
```

matks (approved):
As I am not happy how things turned out (see https://github.com/PrestaShop/ps_facetedsearch/pull/983#issuecomment-1941625243) I happily approve this PR.

Looks like we've had plenty people motivated for introducing the change https://github.com/PrestaShop/PrestaShop/pull/34184 but when it's time to do some work and carry out the needed changes for the changes to be complete https://github.com/PrestaShop/PrestaShop/issues/34578 only me and Hlavtox did some work 🙄 . I am disappointed.

So I'm fine with this PR.

clotairer:
Not sure comments opposing dev secops and opensource dev or a national cliche serve the cause... I'm all days on Slack if you want request me something and not with the same frequency on Github. Neverminds.
Could you please define me a dead line ? I'll try to find a developper for this issue.

jolelievre:
@kpodemski @Hlavtox hmm, you're right introducing dedicated controllers would be cleaner, I'm gonna do that to show the example

clotairer:
@jolelievre Why not crate Command ?
For website in PS 9, already compliant. For previous PS release, the old script could be stay in place and deprecated.

jolelievre:
@clotairer creating command to be run in CLI would be nice, but it's not the same behaviour as the current one which is based on URLs that can be called (I guess it's mainly for merchants who do not have access to cron editing on their server and use external third party services that call the URL as a webhook), so in this PR I only focused on making a similar behaviour

@kpodemski @Hlavtox I created a module front controller to replace these endpoints, I tried to create an Admin one for this but couldn't make it work on develop branch since we now have an automatic CQRF token checking on each request and it requires having a logged session to be validated So external calls are not compatible with CSRF protection which makes sense, so the only solution I found was to put this endpoint on the FO side, it sill integrates its custom token validation system which is the same as the standalone endpoint

The standalone endpoints have been deprecated and should be cleaned in the next major version of this module along with the `.htaccess` file

controllers/front/cron.php
```diff
@@ -0,0 +1,80 @@
...
+{
+    public function __construct()
+    {
+        parent::__construct();
+        $this->ajax = true;
+    }
+
+    public function postProcess()
+    {
+        if (substr(Tools::hash('ps_facetedsearch/index'), 0, 10) != Tools::getValue('token') || !Module::isInstalled('ps_facetedsearch')) {
```
Hlavtox, line 39: I think that `!Module::isInstalled('ps_facetedsearch')` can be removed, the controller should not be callable if the module is not active or installed.

ps_facetedsearch-attribute-indexer.php
```diff
@@ -17,6 +17,14 @@
  * @copyright Since 2007 PrestaShop SA and Contributors
  * @license   https://opensource.org/licenses/AFL-3.0 Academic Free License 3.0 (AFL-3.0)
  */
+
+/*
+ * This standalone endpoint is deprecated, it should not be sued anymore and should be removed along with the
```
matks, line 22:
```suggestion
 * This standalone endpoint is deprecated, it should not be used anymore and should be removed along with the
```
Yes please don't sue it the poor guy

ps_facetedsearch-clear-cache.php
```diff
@@ -17,6 +17,14 @@
  * @copyright Since 2007 PrestaShop SA and Contributors
  * @license   https://opensource.org/licenses/AFL-3.0 Academic Free License 3.0 (AFL-3.0)
  */
+
+/*
+ * This standalone endpoint is deprecated, it should not be sued anymore and should be removed along with the
```
matks, line 22:
```suggestion
 * This standalone endpoint is deprecated, it should not be used anymore and should be removed along with the
```

ps_facetedsearch-price-indexer.php
```diff
@@ -17,6 +17,14 @@
  * @copyright Since 2007 PrestaShop SA and Contributors
  * @license   https://opensource.org/licenses/AFL-3.0 Academic Free License 3.0 (AFL-3.0)
  */
+
+/*
+ * This standalone endpoint is deprecated, it should not be sued anymore and should be removed along with the
```
matks, line 22:
```suggestion
 * This standalone endpoint is deprecated, it should not be used anymore and should be removed along with the
```

ps_facetedsearch.php
```diff
@@ -779,20 +779,20 @@ public function getContent()
...
             'current_url' => Tools::safeOutput(preg_replace('/&deleteFilterTemplate=[0-9]*&id_layered_filter=[0-9]*/', '', $_SERVER['REQUEST_URI'])),
             'id_lang' => $this->getContext()->cookie->id_lang,
-            'token' => substr(Tools::hash('ps_facetedsearch/index'), 0, 10),
+            'token' => $cronToken,
             'base_folder' => urlencode(_PS_ADMIN_DIR_),
-            'price_indexer_url' => $moduleUrl . 'ps_facetedsearch-price-indexer.php' . '?token=' . substr(Tools::hash('ps_facetedsearch/index'), 0, 10),
-            'full_price_indexer_url' => $moduleUrl . 'ps_facetedsearch-price-indexer.php' . '?token=' . substr(Tools::hash('ps_facetedsearch/index'), 0, 10) . '&full=1',
-            'attribute_indexer_url' => $moduleUrl . 'ps_facetedsearch-attribute-indexer.php' . '?token=' . substr(Tools::hash('ps_facetedsearch/index'), 0, 10),
-            'clear_cache_url' => $moduleUrl . 'ps_facetedsearch-clear-cache.php' . '?token=' . substr(Tools::hash('ps_facetedsearch/index'), 0, 10),
+            'price_indexer_url' => $this->context->link->getModuleLink('ps_facetedsearch', 'cron', ['ajax' => true, 'action' => 'indexPrices', 'token' => $cronToken]),
```
matks, line 792: 👍

matks (approved):

jolelievre:
Thanks @matks and @Hlavtox all should be good now

matks (approved):

boherm (approved):

kpodemski (approved):

AureRita:
Hi @jolelievre 

Thank you for your PR, I tested it and it seems to works on develop and also on 8.1.x, as we can see : 

[recording(104).webm](https://github.com/PrestaShop/ps_facetedsearch/assets/100699445/b087ac04-33dd-4159-ae80-29391f492249)

Because the PR seems to works as expected, It's QA :heavy_check_mark: 

Thank you

AureRita (approved):

PR #984, opened 2024-02-13
Hide out of stock products from any filter

msaustral:
| Questions     | Answers
| Description?  | We have add an option on the settings to hide products without stock, this solve the issue https://github.com/PrestaShop/PrestaShop/issues/10018. Also when this option is enable, it disable the Availability filter, if the filter was enabled before activating the option "Hide out of stock", it deletes from table layered_category any availability type register to disable it
| Type?         | improvement 
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | https://github.com/PrestaShop/PrestaShop/issues/10018
| How to test?  | On the setting enable the option: "Hide out of stock"

image

### DIFF
--- a/ps_facetedsearch.php
+++ b/ps_facetedsearch.php
@@ -151,12 +151,9 @@ class Ps_Facetedsearch extends Module implements WidgetInterface
 
     protected function getDefaultFilters()
     {
-        return [
+        $filters = [
             'layered_selection_subcategories' => [
                 'label' => 'Sub-categories filter',
-            ],
-            'layered_selection_stock' => [
-                'label' => 'Product stock filter',
             ],
             'layered_selection_condition' => [
                 'label' => 'Product condition filter',
@@ -176,6 +173,17 @@ class Ps_Facetedsearch extends Module implements WidgetInterface
                 'label' => 'Product extras filter',
             ],
         ];
+        // if we are not showing out of stock products, we do not show the Availability filter
+        if (Configuration::get('PS_LAYERED_FILTER_HIDE_OUT_OF_STOCK') === '0'){
+            $filters = array_merge(
+                $filters,[
+                    'layered_selection_stock' => [
+                        'label' => 'Product stock filter',
+                    ]
+                ]
+            );
+        }
+        return $filters;
     }
 
     public function install()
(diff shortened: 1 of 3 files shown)


msaustral:
also solve issue: https://github.com/PrestaShop/PrestaShop/issues/16952

Hlavtox:
What's the purpose of this filter? You can just click "Available" in Availability filter and it will show you the exact same thing.

gdoleczek:
Hi, what is the progress on this review.
It would be useful on our projects as well.

boo-code:
The stock filter is dropped from `getDefaultFilters()` on every shop that upgrades, rather than only on shops that asked for it.

```php
if (Configuration::get('PS_LAYERED_FILTER_HIDE_OUT_OF_STOCK') === '0'){
```

`PS_LAYERED_FILTER_HIDE_OUT_OF_STOCK` is created only in `install()`, and this PR adds no upgrade script, so a shop that already has the module installed has no row for it. `Configuration::get()` is declared `get($key, $idLang = null, $idShopGroup = null, $idShop = null, $default = false)` and ends with `return $default;` when `hasKey()` misses, so the call returns `false`, not `'0'`. `false === '0'` is false, so `layered_selection_stock` is never merged in and the Availability filter disappears from the filter templates screen — for merchants who never enabled the new setting.

Two things would close it:

- compare on the value rather than its type and spelling. The same file already does this two methods down: `'hide_out_of_stock' => (bool) Configuration::get('PS_LAYERED_FILTER_HIDE_OUT_OF_STOCK')` in `renderAdminMain()`. `if (!Configuration::get('PS_LAYERED_FILTER_HIDE_OUT_OF_STOCK'))` reads the same on a missing key and on a stored `0`.
- add the upgrade script that sets the default, next to the existing `upgrade/upgrade-4.0.4.php`, so the key exists on upgraded shops for the rest of the code that reads it.

Worth noting for whoever reviews the behaviour rather than the code: the `DELETE FROM ps_layered_category WHERE type = "availability"` in `getContent()` is not reversible from the UI. Turning the setting on and off again leaves the availability filter removed from every saved template, and it has to be re-added by hand.

Hlavtox:
Hello @msaustral, you can now implement your behavior easily via `actionFacetedSearchFilters` hook.

```php
public function hookActionFacetedSearchFilters($params) {
  $params['search']->getSearchAdapter()->addOperationsFilter(
    'with_stock_management',
    [[['quantity', [0], '>']]]
  );
}
```

msaustral:
Hi @Hlavtox the idea of opening this in 2024 (WOW), was to converted in a function that comes already with in the native module without need of building another module on top.

Just adding couple of lines you allow salers to have the control of showing or not productos or attributes out sock

PR #990, opened 2024-02-26
Use loose type for Translator

jolelievre:
| Questions     | Answers
| Description?  | Use loose typing in classes as the core uses different Translator override and decoration that mess with the initial type, also with this the module should remain compatible with more PrestaShop versions
| Type?         | bug fix
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | ~
| How to test?  | This PR only needs to be tested with the develop branch from the core, to validate it fixes the bug that prevents finishing the Attributes page However, the next release PR of the module should be tested with older versions like 8.1 or 8.0 or even as low as 1.7.6 since it's the minimum range

### DIFF
--- a/src/Form/Attribute/FormModifier.php
+++ b/src/Form/Attribute/FormModifier.php
@@ -30,17 +30,21 @@ namespace PrestaShop\Module\FacetedSearch\Form\Attribute;
 
 use PrestaShop\Module\FacetedSearch\Constraint\UrlSegment;
 use PrestaShopBundle\Form\Admin\Type\TranslatableType;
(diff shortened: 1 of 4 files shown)


matthieu-rolland (approved):

florine2623 (approved):
Hello @jolelievre ,

Tested on `develop` only.

- CRUD Attributes groups/values 
- Attributes groups feature flag enabled/disabled
- Disabled/enabled filters, edited filter style, filter limits
- FO > Check if products created are well displayed
- Check if the quantity is correct
- Check products in multiple categories

LGTM ✅ 
Thanks!

boherm (approved):

PR #1003, opened 2024-03-27
use new date_new from product table

matthieu-rolland:
| Questions     | Answers
| Description?  | use new date_new from product table 
| Type?         | improvement
| BC breaks?    | yes
| Deprecations? | no
| Fixed ticket? | 
| How to test?  | to be tested with: https://github.com/PrestaShop/PrestaShop/pull/35645

BC Break:
- This is compatible from PS 9 only, it should be for a major release of the module, maybe we should make a branch for a next major ?


jolelievre (changes requested):
Thanks @matthieu-rolland 

the changes are good, but the mi compatibility of the module must be increased, it now requires PrestaShop 9.0 minimum with this modification

Hlavtox:
Great opportunity to kill compatibility with anything pre 9.0.0

PR #1020, opened 2024-05-09
Improve performance by running only one big query

Hlavtox:
| Questions     | Answers
| Description?  | Please be specific when describing the PR. <br> Every detail helps: versions, browser/server configuration, specific module/theme, etc. Feel free to add more information below this table.
| Type?         | bug fix / improvement / new feature / refacto / critical
| BC breaks?    | yes / no
| Deprecations? | yes / no
| Fixed ticket? | Fixes PrestaShop/Prestashop#{issue number here}.
| How to test?  | Please indicate how to best verify that this PR is correct.

### DIFF
--- a/src/Adapter/AbstractAdapter.php
+++ b/src/Adapter/AbstractAdapter.php
@@ -48,10 +48,6 @@ abstract class AbstractAdapter implements InterfaceAdapter
 
     protected $orderDirection = 'DESC';
 
-    protected $limit = 20;
-
-    protected $offset = 0;
-
     /** @var InterfaceAdapter */
     protected $initialPopulation = null;
 
@@ -311,17 +307,6 @@ abstract class AbstractAdapter implements InterfaceAdapter
     public function setOrderDirection($direction)
     {
         $this->orderDirection = $direction === 'desc' ? 'desc' : 'asc';
-
-        return $this;
-    }
-
-    /**
-     * {@inheritdoc}
-     */
-    public function setLimit($limit, $offset = 0)
-    {
-        $this->limit = $limit ? (int) $limit : null;
-        $this->offset = (int) $offset;
 
         return $this;
     }
--- a/src/Adapter/InterfaceAdapter.php
+++ b/src/Adapter/InterfaceAdapter.php
@@ -41,16 +41,6 @@ interface InterfaceAdapter
     public function setOrderDirection($direction);
 
     /**
-     * Set the limit and offset associated with the current search
-     *
-     * @param int|null $limit
-     * @param int $offset
-     *
-     * @return self
-     */
-    public function setLimit($limit, $offset = 0);
-
-    /**
      * Execute the search
      *
      * @return mixed
--- a/src/Adapter/MySQL.php
+++ b/src/Adapter/MySQL.php
@@ -52,7 +52,6 @@ class MySQL extends AbstractAdapter
         $mysqlAdapter = $this->getFilteredSearchAdapter();
         $mysqlAdapter->copyFilters($this);
         $mysqlAdapter->setSelectFields(['price_min', 'MIN(price_min) as min, MAX(price_max) as max']);
-        $mysqlAdapter->setLimit(null);
         $mysqlAdapter->setOrderField('');
 
         $result = $mysqlAdapter->execute();
@@ -134,10 +133,6 @@ class MySQL extends AbstractAdapter
             if ($orderField !== 'p.id_product') {
                 $query .= ', p.id_product DESC';
             }
-        }
-
-        if ($this->limit !== null) {
-            $query .= ' LIMIT ' . $this->offset . ', ' . $this->limit;
         }
 
         return $query;
@@ -588,7 +583,6 @@ class MySQL extends AbstractAdapter
                 $idTmpFilteredProducts = [];
                 $mysqlAdapter = $this->getFilteredSearchAdapter();
                 $mysqlAdapter->addSelectField('id_product');
-                $mysqlAdapter->setLimit(null);
                 $mysqlAdapter->setOrderField('');
                 $mysqlAdapter->addFilter($filterName, $filterValues, $operator);
                 $idProducts = $mysqlAdapter->execute();
@@ -745,7 +739,6 @@ class MySQL extends AbstractAdapter
         $mysqlAdapter = $this->getFilteredSearchAdapter();
         $mysqlAdapter->copyFilters($this);
         $mysqlAdapter->setSelectFields(['MIN(' . $fieldName . ') as min, MAX(' . $fieldName . ') as max']);
-        $mysqlAdapter->setLimit(null);
         $mysqlAdapter->setOrderField('');
 
         $result = $mysqlAdapter->execute();
@@ -778,7 +771,6 @@ class MySQL extends AbstractAdapter
         }
 
         $this->addSelectField('COUNT(DISTINCT p.id_product) c');
-        $this->setLimit(null);
         $this->setOrderField('');
 
         $this->copyOperationsFilters();
@@ -791,8 +783,7 @@ class MySQL extends AbstractAdapter
      */
     public function useFiltersAsInitialPopulation()
     {
-        // Initial population has NO LIMIT and no ORDER BY
-        $this->setLimit(null);
+        // Initial population has no ORDER BY
         $this->setOrderField('');
 
         // We add basic select fields we will need to matter what
--- a/src/Filters/Products.php
+++ b/src/Filters/Products.php
@@ -65,17 +65,13 @@ class Products
         ProductSearchQuery $query,
         array $selectedFilters = []
     ) {
-        // Get pagination
-        $productsPerPage = (int) $query->getResultsPerPage();
-        $page = (int) $query->getPage();
-
         // Load sorting type and direction, validate it and apply fallback if needed
         $orderBy = $query->getSortOrder()->toLegacyOrderBy(false);
         $orderWay = $query->getSortOrder()->toLegacyOrderWay();
         $orderWay = Validate::isOrderWay($orderWay) ? $orderWay : 'ASC';
         $orderBy = Validate::isOrderBy($orderBy) ? $orderBy : 'position';
 
-        $this->searchAdapter->setLimit($productsPerPage, ($page - 1) * $productsPerPage);
+        // Apply it to the filter
         $this->searchAdapter->setOrderField($orderBy);
         $this->searchAdapter->setOrderDirection($orderWay);
 
@@ -87,19 +83,29 @@ class Products
             $this->searchAdapter->addSelectField('price_max');
         }
 
-        $matchingProductList = $this->searchAdapter->execute();
+        // Get full list of matching products
+        $fullProductList = $this->searchAdapter->execute();
 
-        $this->pricePostFiltering($matchingProductList, $selectedFilters);
+        // Count them
+        $totalProductCount = count($fullProductList);
 
-        $nbrProducts = $this->searchAdapter->count();
+        // Get pagination
+        $productsPerPage = (int) $query->getResultsPerPage();
+        $page = (int) $query->getPage();
 
-        if (empty($nbrProducts)) {
-            $matchingProductList = [];
-        }
+        // Cut them down by pagination
+        $finalProductList = array_slice(
+            $fullProductList,
+            ($page - 1) * $productsPerPage,
+            $productsPerPage
+        );
+
+        // And run post filter
+        $this->pricePostFiltering($cutMatchingProductList, $selectedFilters);
 
         return [
-            'products' => $matchingProductList,
-            'count' => $nbrProducts,
+            'products' => $finalProductList,
+            'count' => $totalProductCount,
         ];
     }
 
--- a/tests/php/FacetedSearch/Adapter/MySQLTest.php
+++ b/tests/php/FacetedSearch/Adapter/MySQLTest.php
@@ -73,7 +73,7 @@ class MySQLTest extends MockeryTestCase
     public function testGetEmptyQuery()
     {
         $this->assertEquals(
-            'SELECT  FROM ps_product p ORDER BY p.id_product DESC LIMIT 0, 20',
+            'SELECT  FROM ps_product p ORDER BY p.id_product DESC',
             $this->adapter->getQuery()
         );
     }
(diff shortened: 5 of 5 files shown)


jocel1:
There are big drawbacks with this proposed fix:

- the whole result will be sent by MySQL to PHP, which will increased a lot the network trafic (and the time to receive the data). With cloud database (AWS, GCP, Azure, ...), the trafic bandwidth is something which should be as low as possible
- PHP memory usage will skyrocket if the amount of data returned by the filter is really high (memory limit issue)

Hlavtox:
@jocel1 I don't think so, the real world tests say otherwise.

> the whole result will be sent by MySQL to PHP, which will increased a lot the network trafic (and the time to receive the data). With cloud database (AWS, GCP, Azure, ...), the trafic bandwidth is something which should be as low as possible

The difference with this fix is only 1 ms, for 97 ms saved.

> PHP memory usage will skyrocket if the amount of data returned by the filter is really high (memory limit issue)

You can see on the screenshot that in category with 10000 products, the memory consumption is exactly the same.

SharakPL (approved):

jocel1:
@Hlavtox It's not the number of product in the category which matters, but the number of lines returned by your query. In your example, how many rows are really returned if you execute the query without the LIMIT clause?
About the 1ms vs 97ms you're testing locally, not on a real server where the bandwidth between front servers & the database is limited (and not free)

ShaiMagal:
Tested on huge customer e-shop (PS 8.1.5).

BEFORE:
BEFORE

AFTER:
AFTER

Hlavtox:
@jocel1 

> @Hlavtox It's not the number of product in the category which matters, but the number of lines returned by your query. In your example, how many rows are really returned if you execute the query without the LIMIT clause?

The amount of returned rows (4 columns) are equal to number of products found.

> About the 1ms vs 97ms you're testing locally, not on a real server where the bandwidth between front servers & the database is limited (and not free)

I tested it on real server in production. All the shops where we tried this show a massive improvement in performance.

jocel1:
> @jocel1
> 
> > @Hlavtox It's not the number of product in the category which matters, but the number of lines returned by your query. In your example, how many rows are really returned if you execute the query without the LIMIT clause?
> 
> The amount of returned rows (4 columns) are equal to number of products found.
> 

Yes but how many rows? 100, 1000, 10000, 100000? Some websites have facetedsearch which returns *a lot* of products / rows. 
The impact in both bandwidth and memory is limited because it only returns ids but we should be careful to not overwhelm the database.

ShaiMagal (approved):

ShaiMagal:
@jocel1 @Hlavtox my real test conditions of my post before:
More than 100 000 products.
Approximatelly 2 000 categories.
It is very busy eshop, but of course i will be happy if more people will test it.

This PR can be gamechanger especially for BIG eshops.

I was "attacking" @Hlavtox about this PR about 2 years already :D so i am happy, it's finally here :)

I hope there will be not problem and PR will be merged to the next release.

Our customers can upgrade module then and benefit from this.

We have a lot of customers with complain on performance of this module, so i will happy update module for all of them :)

jolelievre (dismissed):
I just block this one for no until a consensus is found
I just don't want this one to be merged by mistake or without QA

Hlavtox:
@jocel1 

> Yes but how many rows? 100, 1000, 10000, 100000? Some websites have facetedsearch which returns a lot of products / rows.
> The impact in both bandwidth and memory is limited because it only returns ids but we should be careful to not overwhelm the database.

As I said, the amount of rows returned by the final result is equal to the amount of products.

This query with LIMIT clause takes 183 ms among 10 measurements. Returns 20 items.
```sql
SELECT SQL_NO_CACHE p.id_product FROM (
SELECT p.id_product, p.id_manufacturer, SUM(sa.quantity) as quantity, p.condition, p.weight, p.price, psales.quantity as sales, p.on_sale, p.date_add FROM ps_product p
LEFT JOIN ps_product_attribute pa ON (p.id_product = pa.id_product)
LEFT JOIN ps_product_attribute_combination pac ON (pa.id_product_attribute = pac.id_product_attribute)
LEFT JOIN ps_stock_available sa ON (p.id_product = sa.id_product AND IFNULL(pac.id_product_attribute, 0) = sa.id_product_attribute AND sa.id_shop = 1  AND sa.id_shop_group = 0 )
LEFT JOIN ps_product_sale psales ON (psales.id_product = p.id_product)
INNER JOIN ps_product_shop ps ON (p.id_product = ps.id_product AND ps.id_shop = 1 AND ps.active = TRUE)
INNER JOIN ps_category_product cp ON (p.id_product = cp.id_product)
INNER JOIN ps_category c ON (cp.id_category = c.id_category AND c.active=1)
LEFT JOIN ps_category_group cg ON (cg.id_category = c.id_category)
WHERE ps.id_shop='1' AND ps.visibility IN ('both', 'catalog') AND cg.id_group='1' AND c.nleft>=3 AND c.nright<=28 GROUP BY p.id_product
) p GROUP BY p.id_product ORDER BY p.date_add DESC, p.id_product DESC LIMIT 0, 20
```

When you remove the LIMIT clause, it takes virtually the same time - 186 ms among 10 measurements. Returns 361 items.
```sql
SELECT SQL_NO_CACHE p.id_product FROM (
SELECT p.id_product, p.id_manufacturer, SUM(sa.quantity) as quantity, p.condition, p.weight, p.price, psales.quantity as sales, p.on_sale, p.date_add FROM ps_product p
LEFT JOIN ps_product_attribute pa ON (p.id_product = pa.id_product)
LEFT JOIN ps_product_attribute_combination pac ON (pa.id_product_attribute = pac.id_product_attribute)
LEFT JOIN ps_stock_available sa ON (p.id_product = sa.id_product AND IFNULL(pac.id_product_attribute, 0) = sa.id_product_attribute AND sa.id_shop = 1  AND sa.id_shop_group = 0 )
LEFT JOIN ps_product_sale psales ON (psales.id_product = p.id_product)
INNER JOIN ps_product_shop ps ON (p.id_product = ps.id_product AND ps.id_shop = 1 AND ps.active = TRUE)
INNER JOIN ps_category_product cp ON (p.id_product = cp.id_product)
INNER JOIN ps_category c ON (cp.id_category = c.id_category AND c.active=1)
LEFT JOIN ps_category_group cg ON (cg.id_category = c.id_category)
WHERE ps.id_shop='1' AND ps.visibility IN ('both', 'catalog') AND cg.id_group='1' AND c.nleft>=3 AND c.nright<=28 GROUP BY p.id_product
) p GROUP BY p.id_product ORDER BY p.date_add DESC, p.id_product DESC
```

LIMIT clause in SQL has virtually no performance benefit, because the database has to execute the query fully anyway, and then just cuts the final processed result.

So the only overhead you are adding is maximum single digit ms overhead for the transfer, but you are saving an insane amount of time for the second COUNT query.

TLDR - in all cases, you are saving performance, virtually cutting the load time in half.

jocel1:
@Hlavtox That's my point, your second query without the limit clause doesn't return a lot of rows (only 361 lines).
That means your facetedsearch applies a quite restrictive filter. I want to see the impact of a faceted filter which would return a lot of products without the LIMIT clause. (I have some customer where the facetedsearch returns 300000+ products)

As for your statement "LIMIT clause in SQL has virtually no performance benefit, because the database has to execute the query fully anyway, and then just cuts the final processed result." that's not true.
MySQL has a lot of mecanism to stop scanning the rows earlier, for example if the results are returned in the index order (ie no filesorting). 
That's not the case of this query, but you couldn't generalize that the LIMIT cause has no benefit. 
And also don't forget about the data transfer time / cost, which in usually not cheap at all with text fields (here once again it's not that expensive because we are returning IDs, but in general that's not the case).

Also MySQL is able to apply some optimizations on COUNT(), here we cannot apply those optimization because of the derived table.

It would be perhaps interesting to try to rewrite the second query to remove the useless derived table.

More generally, MySQL is not a good candidate for facetedsearch because its index cannot optimized all the possible combinations, that's why the performances are so poor.

Hlavtox:
@matthieu-rolland @jocel1 Maybe MySQL does apply some performance optimizations in case of LIMIT clauses, I don't doubt that.

But the main point here is, that the COUNT query is taking exactly the same time as the select one.

```sql
SELECT SQL_NO_CACHE COUNT(DISTINCT p.id_product) c FROM (
SELECT p.id_product, p.id_manufacturer, SUM(sa.quantity) as quantity, p.condition, p.weight, p.price, psales.quantity as sales, p.on_sale, p.date_add FROM ps_product p
LEFT JOIN ps_product_attribute pa ON (p.id_product = pa.id_product)
LEFT JOIN ps_product_attribute_combination pac ON (pa.id_product_attribute = pac.id_product_attribute)
LEFT JOIN ps_stock_available sa ON (p.id_product = sa.id_product AND IFNULL(pac.id_product_attribute, 0) = sa.id_product_attribute AND sa.id_shop = 1  AND sa.id_shop_group = 0 )
LEFT JOIN ps_product_sale psales ON (psales.id_product = p.id_product)
INNER JOIN ps_product_shop ps ON (p.id_product = ps.id_product AND ps.id_shop = 1 AND ps.active = TRUE)
INNER JOIN ps_category_product cp ON (p.id_product = cp.id_product)
INNER JOIN ps_category c ON (cp.id_category = c.id_category AND c.active=1)
LEFT JOIN ps_category_group cg ON (cg.id_category = c.id_category)
WHERE ps.id_shop='1' AND ps.visibility IN ('both', 'catalog') AND cg.id_group='1' AND c.nleft>=3 AND c.nright<=28 GROUP BY p.id_product
) p;
```

> That being said it's probable that over all your pr improves performance

Yes, by about 50%. :-)

> I'm just not sure how bad it can be for people having small MySQL servers and heavy usage

It reduces the database load by two times, because the database doesn't have to execute the COUNT query.

matthieu-rolland (approved):
Ok, @Hlavtox convinced me, this PR looks got to me 👍 

By the way we have someone in the QA team dedicated to performance subjects, it's definitely the kind of PR he could be interested to test

kpodemski (approved):
I was skeptical, but considering the array consists only of product IDs, it seems that PHP shouldn't have to struggle too much to work on it and perform array_slice. I've performed some tests, and even on 300k randomly generated numbers array_slice acts ok

jocel1:
> I was skeptical, but considering the array consists only of product IDs, it seems that PHP shouldn't have to struggle too much to work on it and perform array_slice. I've performed some tests, and even on 300k randomly generated numbers array_slice acts ok

Yes that's really because it only returns ids it could worth it.

matthieu-rolland:
@kpodemski & @Hlavtox 

About the array_slice: https://stackoverflow.com/questions/6153307/array-slice-with-negative-offset-on-huge-arrays

Apparently, using a for loop instead of an array slice, on 1 million elements, this guy goes from 2.5689ms to 0.0031ms

Could be an interesting optim for edge cases with huge results ?

Hlavtox:
Not sure this micro optimizing is worth it, when we have things like this added in every version - https://github.com/PrestaShop/PrestaShop/blob/9fe825d70e7f19a9fea209c7ccafa788e037046a/classes/Pack.php#L88 :-))))

Hlavtox:
@jocel1 For curiosity, I tested how long it takes to select and retrieve 100k rows - amateur way. :-)))

I have `93092` rows in `ps_accessory`.
I call `Db::getInstance()->executeS('SELECT id_product_1 FROM ps_accessory');`
The query takes about 14 ms to execute and transfer, measured on PHP side.

So if you have 100k products on your shop, it would take right now:
- `2000 ms` to select it with limit
- `2000 ms` to count it

And after the change:
- At most `2014 ms` to select it without a limit. And I am sure that 99% of that 14 ms is the actual select time, not the transfer. I don't know how to measure it.

It can just never be slower than right now.

matthieu-rolland:
> * At most `2014 ms` to select it without a limit. And I am sure that 99% of that 14 ms is the actual select time, not the transfer. I don't know how to measure it.
> 

@jocel1 when you talked about data transfer between cloud servers, did you talk about performance or did you have the price increase in mind ?

jocel1:
@matthieu-rolland price & bandwidth usages mostly, although there could be a small impact on performances depending of the speed of the connection between front servers and db servers.

Hlavtox:
@jocel1 Is it even possible to run Prestashop (or any app) on architecture, where the connection takes more than virtually nothing? There is soooo many queries.

jocel1:
@Hlavtox Yes it's possible, I'm running big clients on AWS or GCP. Decathlon for example is using Prestashop 8 with this kind of setup, so we want to avoid degrading the performances for this kind of customers.

Hlavtox:
@jocel1 Hmmm hm hmmm.

And are these kinds of big shops even using faceted search anymore? We have about 15k products on the biggest store and I am actively searching for ways to go on elasticsearch-like solution.

Would it be possible for you to test it on some big client's dev environment?

jocel1:
> @jocel1 Hmmm hm hmmm.
> 
> And are these kinds of big shops even using faceted search anymore? We have about 15k products on the biggest store and I am actively searching for ways to go on elasticsearch-like solution.
> 
> Would it be possible for you to test it on some big client's dev environment?

For some of them, yes! I'll give it a try

jocel1:
The result on a website with a lot of products (1M+) and which have the faceted search returning a lot of rows, with AWS aurora (based on MySQL 8) 

image

ShaiMagal:
BTW:
This PR can fix this issue:
https://github.com/PrestaShop/PrestaShop/issues/35524

kpodemski:
@jocel1 I saw your comment with not so great perf results, but I think you deleted it. Do you plan to re-run the tests or....?

jocel1:
@kpodemski yes there were some other optimisations in the module removed by the module upgrade, so I put back the original optimized module and I'm planning to patch it with this PR changes.

Hlavtox:
@jocel1 It would be great if you could share some perf mods in other places, if they are not client specific. 😜

jocel1:
> @jocel1 It would be great if you could share some perf mods in other places, if they are not client specific. 😜

one easy optimisation, but only useful with master / slave setup is to use _PS_USE_SQL_SLAVE_ in getDatabase() of MySQL.php

jocel1:
The impact on the DB load is visible (after May 17th here):

image

No significant changes on the DB global bandwidth.
Benchmark from an AWS Aurora 3.06.0 DB instance based on MySQL 8

So this looks good to me

Hlavtox:
All ready, ping @PrestaShop/qa-functional

boherm (approved):

ShaiMagal:
Tested on 3 websites, no negative impact until now. So I believe it will be without problems, because performance benefit is **HUGE**, really.

i hope a new version will be released soon so customers can update it

Thank you @Hlavtox

paulnoelcholot:
Hello @Hlavtox,

I tested your PR and it's good for me! :tada: 

Thanks!

paulnoelcholot (approved):

elyesnaouar:
The improvement is checked and the performance gain is clearly visible.

[PR1020_Big Query Approuvement .pdf](https://github.com/user-attachments/files/15840522/PR1020_Big.Query.Approuvement.pdf)

PR #1021, opened 2024-05-09
Make backport return items ordered by position

Hlavtox:
| Questions     | Answers
| Description?  | In preparation to finalize search controller support, our backport needs to return items in proper position, which we can later use. 
| Type?         | refacto
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | 
| How to test?  | This is an experimental feature and does not need testing.

### DIFF
--- a/src/Product/CoreSearchBackport.php
+++ b/src/Product/CoreSearchBackport.php
@@ -324,6 +324,7 @@ class CoreSearchBackport
         $context = Context::getContext();
         $db = Db::getInstance(_PS_USE_SQL_SLAVE_);
 
+        $scoreArray = [];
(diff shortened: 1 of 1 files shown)


ShaiMagal (approved):

Hlavtox:
@kpodemski Can I merge this as this concerns only the experimental feature implementation?

florine2623 (approved):
Hello @Hlavtox ,

Tested the experimental feature for search in FO ^^
Checked if items were ordered by position. They are order by ID DESC automatically. When edited the position, the search result will update with the current position.

Works as expected ✅ 
Thanks!

PR #1083, opened 2024-09-27
Use feature value position for ordering

ks129:
| Questions     | Answers
| Description?  | Use feature value new position field for ordering of feature values. Part of PrestaShop/PrestaShop#37042
| Type?         | new feature
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | part of PrestaShop/PrestaShop#35499 and PrestaShop/PrestaShop#32267
| How to test?  | If used together with parent PR, it should show feature values in faceted search in the order provided in BO.

### DIFF
--- a/ps_facetedsearch.php
+++ b/ps_facetedsearch.php
@@ -232,6 +232,8 @@ class Ps_Facetedsearch extends Module implements WidgetInterface
             }
         }
 
+        Configuration::updateValue('PS_LAYERED_FILTER_FEATURE_VALUES_USE_POSITION', 0);
+
         return true;
     }
 
@@ -247,6 +249,7 @@ class Ps_Facetedsearch extends Module implements WidgetInterface
         Configuration::deleteByName('PS_LAYERED_FILTER_PRICE_ROUNDING');
         Configuration::deleteByName('PS_LAYERED_FILTER_SHOW_OUT_OF_STOCK_LAST');
         Configuration::deleteByName('PS_LAYERED_FILTER_BY_DEFAULT_CATEGORY');
+        Configuration::deleteByName('PS_LAYERED_FILTER_FEATURE_VALUES_USE_POSITION');
 
         $this->getDatabase()->execute('DROP TABLE IF EXISTS ' . _DB_PREFIX_ . 'layered_category');
         $this->getDatabase()->execute('DROP TABLE IF EXISTS ' . _DB_PREFIX_ . 'layered_filter');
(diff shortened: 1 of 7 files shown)


Hlavtox (changes requested):
@ks129 I like this, but please keep the option to sort the features alphabetically.

I have only one client that sells clothes.

For all other clients, it will be a pain in the ass when they would have to sort the features manually. Alphabetical sorting is OK for most of the stores.

ps-jarvis:

kpodemski:
> [!NOTE]

Hi @ks129, sorry this waited so long. @Hlavtox asked to keep alphabetical sorting as an option and you did that, so the request is answered. I rebased the branch on current `dev` and pushed a few fixes on top, because the feature did not actually work end to end.

> [!IMPORTANT]
> Your branch was rebased and force updated. Before you push again:
> ```
> git fetch origin && git reset --hard origin/feature-value-position
> ```

## The setting had no effect on the front office

The `ORDER BY` you added in `DataAccessor` never reached the shop. The feature block values are keyed by the facet count results, not by that query, and then two sorts run after it:

1. `Block::sortFeatureBlock()` calls `natcasesort` on the names
2. `Converter` calls `usort` with `sortFiltersByLabel` on the same names

So the values were alphabetical either way. Turning the setting on and off gave an identical page.

The fix carries the position onto the filter and orders by it in `Converter`, which is the last sort and therefore the one that decides. Doing it there also keeps working when a filter has a display limit, where values are first cut by magnitude.

## It would have broken PrestaShop 8

`feature_value.position` arrives with PrestaShop/PrestaShop#37042, milestone 9.0.0. This module still declares `min => 8.2.0` and CI runs against 8.2.x, so `ORDER BY v.position` there is an unknown column. The ordering is now guarded by a version check, and the setting is hidden in the back office on cores that do not have the column, so nobody is offered a switch that cannot work.

## Other changes

- `use Configuration;` was ordered before `use Combination;`, which fails PHP-CS-Fixer.
- The new option had no description while every other one has, so I added one.
- The template block was indented with spaces instead of a tab.
- `BlockTest` and `DataAccessorTest` needed the new configuration key, otherwise the suite errors.

CI is green now, 117 tests.

## QA

Back office, the new setting with its description:

qa-config

The Composition feature values are stored in this order: Polyester (0), Wool (1), Ceramic (2), Cotton (3), Recycled cardboard (4), Matt paper (5).

Setting off, values sorted alphabetically, which is the current behaviour and what @Hlavtox wanted kept:

qa-alphabetical

Setting on, values in the order the merchant arranged them. Wool is missing because no product in this category has it:

qa-position

kpodemski (approved):

kpodemski:
That's one for you @Hlavtox

Hlavtox (approved):
Good!

PR #1093, opened 2024-10-14
Performance improvements

chrisai-dev:
| Questions     | Answers
| Description?  | Improved both cold and hot query performance by 2-15x
| Type?         |  improvement
| BC breaks?    | no
| Deprecations? |  no
| Fixed ticket? | Fixes PrestaShop/Prestashop#{issue number here}.
| How to test?  | Please indicate how to best verify that this PR is correct.

Changes:

- id_shop filter will be only added if multishop feature is enabled
- product visibility filter is removed from the base query and added to the outer
- category_product table join on the outer query was removed
- removed all unused selects/joins from base query such as id_manufacture, price, on_sale etc. and added them as joins to $filterToTableMapping.

Performance:
There are the execution times for the queries generated by the module. 
Cold means the faceted blocks are not yet cached, and warm means the blocks are cached, only the base query is running.

- category 1:
  cold: 412ms -> 74ms
  warm: 31ms -> 2ms
- category 2:
  cold: 673ms -> 258ms
  warm: 46ms -> 14ms
- category 3:
  cold: 506ms -> 105ms
  warm: 36ms -> 5ms
- category 4:
  cold: 81ms -> 23ms
  warm: 4ms -> 1ms

Test setup is a dev machine with ryzen 7700x, nvme ssd, 32gb ram, I expect better improvements on weaker configs.

### DIFF
--- a/src/Product/Search.php
+++ b/src/Product/Search.php
@@ -329,11 +329,10 @@ class Search
      */
     private function addCommonFilters()
     {
-        // Setting proper shop
-        $this->getSearchAdapter()->addFilter('id_shop', [(int) $this->context->shop->id]);
-
-        // Visibility of a product must be in catalog or both (search & catalog)
-        $this->addFilter('visibility', ['both', 'catalog']);
+        // Setting proper shop if multishop is enabled
+        if (Configuration::get('PS_MULTISHOP_FEATURE_ACTIVE')) {
+            $this->getSearchAdapter()->addFilter('id_shop', [(int) $this->context->shop->id]);
+        }
 
         // User must belong to one of the groups that can access the product
         // (Actually it's categories that define access to a product, user must have access to at least
--- a/src/Adapter/MySQL.php
+++ b/src/Adapter/MySQL.php
@@ -95,6 +95,11 @@ class MySQL extends AbstractAdapter
         // Prepare mapping for joined tables
         $filterToTableMapping = $this->getFieldMapping();
 
+        // visibility filter is added only to the outer query
+        if ($this->getInitialPopulation() !== null) {
+            $this->addFilter('visibility', ['both', 'catalog']);
+        }
+
         // Process and generate all fields for the SQL query below
         $orderField = $this->computeOrderByField($filterToTableMapping);
         $selectFields = $this->computeSelectFields($filterToTableMapping);
@@ -323,6 +328,41 @@ class MySQL extends AbstractAdapter
                     (sp.to = \'0000-00-00 00:00:00\' OR \'' . date('Y-m-d H:i:s') . '\' <= sp.to) 
                 )',
                 'joinType' => self::LEFT_JOIN,
+            ],
+            'id_manufacturer' => [
+                'tableName' => 'product',
+                'tableAlias' => 'mid',
+                'fieldName' => 'id_manufacturer',
+                'joinCondition' => '(p.id_product = mid.id_product)',
+                'joinType' => self::INNER_JOIN,
+            ],
+            'on_sale' => [
+                'tableName' => 'product',
+                'tableAlias' => 'ons',
+                'fieldName' => 'on_sale',
+                'joinCondition' => '(p.id_product = ons.id_product)',
+                'joinType' => self::INNER_JOIN,
+            ],
+            'date_add' => [
+                'tableName' => 'product',
+                'tableAlias' => 'dadd',
+                'fieldName' => 'date_add',
+                'joinCondition' => '(p.id_product = dadd.id_product)',
+                'joinType' => self::INNER_JOIN,
+            ],
+            'condition' => [
+                'tableName' => 'product',
+                'tableAlias' => 'pc',
+                'fieldName' => 'condition',
+                'joinCondition' => '(p.id_product = pc.id_product)',
+                'joinType' => self::INNER_JOIN,
+            ],
+            'price' => [
+                'tableName' => 'product',
+                'tableAlias' => 'prc',
+                'fieldName' => 'price',
+                'joinCondition' => '(p.id_product = prc.id_product)',
+                'joinType' => self::INNER_JOIN,
             ],
         ];
 
@@ -652,7 +692,8 @@ class MySQL extends AbstractAdapter
 
         $this->addJoinList($joinList, $this->getGroupFields()->getKeys(), $filterToTableMapping);
 
-        if (array_key_exists($this->getOrderField(), $filterToTableMapping)) {
+        // position field is part of the base query, no need to rejoin category_product
+        if (array_key_exists($this->getOrderField(), $filterToTableMapping) && $this->getOrderField() != 'position') {
             $joinMapping = $filterToTableMapping[$this->getOrderField()];
             $this->addJoinConditions($joinList, $joinMapping, $filterToTableMapping);
         }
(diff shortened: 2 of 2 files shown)


kpodemski:
Hello @chrisai-dev 

Thank you for your pull request. Very interesting results. Could you share more details about the number of products in a tested category, features, etc.?

Also, it would be great to understand better this part:
> removed all unused selects/joins from base query such as id_manufacture, price, on_sale etc. and added them as joins to $filterToTableMapping.

The rest is fully understandable.

SharakPL:
> id_shop filter will be only added if multishop feature is enabled

I'm not sure that's wise. What if there are still multishop products in the database, but the user decided to disable multishop temporarily? Won't this make the query include the products from all shops?

chrisai-dev:
> > id_shop filter will be only added if multishop feature is enabled
> 
> I'm not sure that's wise. What if there are still multishop products in the database, but the user decided to disable multishop temporarily? Won't this make the query include the products from all shops?

@SharakPL, Once you have enabled multishop, and uploaded a product, you cannot disable the feature. No need to add this filter condition for a single store.



> Hello @chrisai-dev
> 
> Thank you for your pull request. Very interesting results. Could you share more details about the number of products in a tested category, features, etc.?
> 
> Also, it would be great to understand better this part:
> 
> > removed all unused selects/joins from base query such as id_manufacture, price, on_sale etc. and added them as joins to $filterToTableMapping.
> 
> The rest is fully understandable.

Hi @kpodemski,

So it's been a while and I've deleted the test results, but, the store roughly had 4k products, 1k categories, the ps_category_product table had about 22000 rows. I tested very large categories with about 1k product, mid sized ones with 100ish, and a small one with 10.

https://github.com/PrestaShop/ps_facetedsearch/blob/79d694e1dcda9a7745677b8827188b1b3fed0bce/src/Adapter/MySQL.php#L789-L802

So this is the part where the fields for the base query are declared. The only field that is "needed no matter what", is id_product, the rest is only needed for specific queries. But these fields had no mappings to tables (https://github.com/PrestaShop/ps_facetedsearch/blob/79d694e1dcda9a7745677b8827188b1b3fed0bce/src/Adapter/MySQL.php#L154)
since they were part of the base query, I added them. 

This is the function that computes the joins based on the query
https://github.com/PrestaShop/ps_facetedsearch/blob/79d694e1dcda9a7745677b8827188b1b3fed0bce/src/Adapter/MySQL.php#L619

I only tested on 8.1.x, don't know if this will affect compatibility with  <= 1.7.

SharakPL:
> > > id_shop filter will be only added if multishop feature is enabled
> > 
> > 
> > I'm not sure that's wise. What if there are still multishop products in the database, but the user decided to disable multishop temporarily? Won't this make the query include the products from all shops?
> 
> @SharakPL, Once you have enabled multishop, and uploaded a product, you cannot disable the feature. No need to add this filter condition for a single store.

Not sure about PS 8 & 9, but you can disable multishop on 1.7 without problems on Preferences main page. All shops after that work just like before. The only difference is you can't manage additional shops, only the default one.

By removing `id_shop` from filters you'll force the default shop to load all the products from all shops.

src/Product/Search.php
```diff
@@ -329,11 +329,10 @@ private function addSearchFilters($selectedFilters)
...
     {
-        // Setting proper shop
-        $this->getSearchAdapter()->addFilter('id_shop', [(int) $this->context->shop->id]);
-
-        // Visibility of a product must be in catalog or both (search & catalog)
-        $this->addFilter('visibility', ['both', 'catalog']);
+        // Setting proper shop if multishop is enabled
+        if (Configuration::get('PS_MULTISHOP_FEATURE_ACTIVE')) {
+            $this->getSearchAdapter()->addFilter('id_shop', [(int) $this->context->shop->id]);
+        }
```
SharakPL, line 335:
```suggestion
        if (Shop::isFeatureActive()) {
            $this->getSearchAdapter()->addFilter('id_shop', [(int) $this->context->shop->id]);
        }
```

kpodemski:
> [!NOTE]
> Semi-automated review: the benchmarks below were run by Claude Code against a real PrestaShop instance, and every number was checked and signed off by me before posting. A human is in the loop and answering — reply here as you normally would.

Hi @chrisai-dev — first of all, I'm sorry this sat for so long. You opened it in October 2024 with a clear diagnosis and real measurements, and it deserved a much faster answer than this. Thank you for the patience.

The good news is that your analysis was right, and the module ended up going exactly where you pointed. The base query really was dragging unused columns and joins around, and that has since been fixed on `dev` by @Hlavtox in #1303, which measured about a **5× improvement** on a large catalogue. Comparing your branch against `dev` today, the three main ideas are already there:

| your PR | `dev` today |
| --- | --- |
| strip the unused selects from the base query | `setSelectFields(['id_product'])` |
| re-add those fields as mappings | `addRequiredInitialPopulationFields()` — adds only what the outer query asks for |
| `&& getOrderField() != 'position'` | `requiresMappedTable()`, the same idea generalised |

So I rebased your branch onto current `dev` and measured what the remaining delta is worth.

### Setup

PrestaShop 9.1.5 · PHP 8.3 · MariaDB 11.4 · 100 000 products, 400 000 combinations, 4 166-product category. Best-of-3, two rounds with the order reversed.

### Results

| page | `dev` | this PR | change |
| --- | ---: | ---: | ---: |
| 100 000-product category | 16.52s | 14.88s | **−10%** |
| 4 166-product category | 1.257s | 1.392s | **+11%** |
| category, sorted by price | 1.149s | 1.393s | **+21%** |
| category, sorted by sales | 1.191s | 1.437s | **+21%** |
| control: product page | 0.165s | 0.166s | +0.9% |
| control: `/search` | 0.050s | 0.050s | +0.6% |

Controls stayed flat, and both rounds agreed, so the differences are real. On top of `dev` it now wins only on the pathological 100 k category and costs 10–20 % on the pages a normal shop actually serves.

### Where that comes from

Splitting the change in two and measuring each half separately: the multishop guard on `id_shop` is **free** (+1.5 %, −0.7 %, 0.0 % — noise). Moving the visibility filter to the outer query carries all of it, and on its own is worse than the full PR:

| page | visibility move only | `id_shop` guard only |
| --- | ---: | ---: |
| sorted by price | +25 … +30% | +1.6% |
| sorted by sales | +18 … +24% | −4.4 … −0.7% |
| category | +13 … +15% | −2.7 … 0.0% |

The reason is that the filter does not really move. Same page, same 481 statements:

| | `dev` | visibility moved out |
| --- | ---: | ---: |
| total SQL time | 0.945s | 1.247s |
| rows examined | 2 055 245 | 2 522 015 |
| `product_shop` joined in the inner query | 22 | 22 |
| `product_shop` joined in the **outer** query | 0 | 8 |

The inner join is still there — `id_shop` keeps it — and the outer query now joins `product_shop` as well, on eight facet and count queries. It is an extra join rather than a relocated one.

### Correctness

For what it's worth, the branch is behaviourally clean: 162 comparable cases (2 categories × 8 filter sets × 8 sort orders × out-of-stock-last on/off) showed **zero** differences in product counts, product order and facet counts. Nothing here is a criticism of the change itself — it simply lands on a very different `dev` than the one you wrote it against.

Given that, I'm going to close this as superseded by #1303. Please don't read that as the work being wasted: you identified the right bottleneck a year before it was fixed, and the fix that shipped is the one you proposed. I'd genuinely welcome more PRs from you.

Caveat, as always: one engine and one data shape. If you have a setup where you see this differently, share the `EXPLAIN` and I'll happily re-open and re-measure.

PR #1123, opened 2025-01-29
Fixed incorrect language condition in LEFT JOIN (liaglv.id_lang instead of agl.id_lang)

abdullacm:
| Questions     | Answers
| Description?  | In the current implementation of the SQL query, the language condition in the LEFT JOIN for the ps_layered_indexable_attribute_group_lang_value table is incorrectly applied to the agl (attribute group language) table. This causes the language condition to be applied to the wrong table, which leads to incorrect results in multi-language environments. 
| Type?         | bug fix
| BC breaks?    | no
| Deprecations? |  no
| How to test?  | In the PrestaShop back office, set the url_name (friendly URL) for the attribute (e.g., size) in multiple languages. On the frontend, switch between languages and apply the filter. Check that the friendly URL in the filter matches the correct language-specific url_name as configured in the backend.

Screenshot 2025-01-29 at 5 37 57 PM

Screenshot 2025-01-29 at 5 39 05 PM

The query is expected to return the url_name (e.g., storrelse) for language ID 3, but it is currently returning the url_name (e.g., size) for language ID 1. This results in incorrect URL names being used for filters in multi-language environments.

### DIFF
--- a/src/Filters/DataAccessor.php
+++ b/src/Filters/DataAccessor.php
@@ -133,7 +133,7 @@ class DataAccessor
                 'LEFT JOIN `' . _DB_PREFIX_ . 'layered_indexable_attribute_group` liag ' .
                 'ON (ag.`id_attribute_group` = liag.`id_attribute_group`) ' .
                 'LEFT JOIN `' . _DB_PREFIX_ . 'layered_indexable_attribute_group_lang_value` AS liaglv ' .
-                'ON (ag.`id_attribute_group` = liaglv.`id_attribute_group` AND agl.`id_lang` = ' . (int) $idLang . ') ' .
+                'ON (ag.`id_attribute_group` = liaglv.`id_attribute_group` AND liaglv.`id_lang` = ' . (int) $idLang . ') ' .
                 'GROUP BY ag.id_attribute_group ORDER BY ag.`position` ASC'
             );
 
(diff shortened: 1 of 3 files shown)


ps-jarvis:
Hello @abdullacm!

SharakPL:
Hello @abdullacm and thanks for the PR. Nice catch!
Looks like there's a test that also needs fixing https://github.com/PrestaShop/ps_facetedsearch/blob/e3d8f332b60a116a4e4c054a746a4df201f02621/tests/php/FacetedSearch/Filters/BlockTest.php#L1078

Hlavtox (dismissed):

paulnoelcholot (commented):
Hello @abdullacm,

In order to be able to test your PR, it is necessary that an issue is created before that.

Thanks for doing that 🙏

kpodemski (approved):
> [!NOTE]

Good catch, and the fix is correct. With `url_name` set per language (`size` / `taille`) and the shop browsed in French:

| | filter URL |
| --- | --- |
| before | `size-S`, `size-M`, `size-L` |
| after | `taille-S`, `taille-M`, `taille-L` |

The facet label was always right ("Taille") — only the URL segment took the wrong language, exactly as described. The three sibling joins (`lialv`, `liflv`, `lifvlv`) already restrict on their own alias, so this was the lone outlier.

@SharakPL was right about the test, and that is why this stalled: the suite is green at the base and goes to 2 errors with the fix, because `BlockTest` hard-codes the old SQL.

I rebased onto `dev` and pushed a commit that updates that stale expectation and adds `DataAccessorTest`, which pins the invariant for all four `layered_indexable_*_lang_value` joins so this typo can't come back in any of them. 116 tests / 1016 assertions green; reverting the fix fails the new test on the attribute-groups data set only.

@abdullacm your branch was force-updated by the rebase — `git fetch origin && git reset --hard origin/fix/incorrect-lang-condition-left-join` before you push again. No issue needed, the reproduction steps here are clear enough.

LGTM.

PR #1124, opened 2025-02-08
Fix: Price index limited to 100 product

### LINKED ISSUE #37918: Price index limited to 100 product
### Prerequisites

- [x] I have already [searched in existing issues](https://github.com/PrestaShop/PrestaShop/issues?q=is%3Aissue+label%3ABug) and found no previous report of this bug.

Price indexing loop only one time in indexPricesUnbreakable.
The second times (and next), we compare $cursor value which is the last productId indexed with the number of products to index ($nbProducts).

File : ps_facetedsearch method IndexPrice.

### Expected behavior

Compare $cursor with productId and not product count

### Steps to reproduce

Launch full index price with product id > 100 in the first loop/

8.1.7

7.2

ps_facetedsearch 3.16.1

Hastone & Ten

Codencode:
| Questions     | Answers
| Description?  | see: https://github.com/PrestaShop/PrestaShop/issues/37918
| Type?         | bug fix
| BC breaks?    | no
| Deprecations? |  no
| Fixed ticket? | Fixes PrestaShop/Prestashop#37918
| How to test?  | see: https://github.com/PrestaShop/PrestaShop/issues/37918

### DIFF
--- a/ps_facetedsearch.php
+++ b/ps_facetedsearch.php
@@ -1554,9 +1554,22 @@ VALUES(' . $last_id . ', ' . (int) $idShop . ')');
                 'INNER JOIN `' . _DB_PREFIX_ . 'product_shop` ps ' .
                 'ON (ps.`id_product` = p.`id_product` AND ps.`active` = 1 AND ps.`visibility` IN ("both", "catalog"))'
             );
+            $maxProductId = (int) $this->getDatabase()->getValue(
+                'SELECT max(DISTINCT p.`id_product`) ' .
+                'FROM ' . _DB_PREFIX_ . 'product p ' .
+                'INNER JOIN `' . _DB_PREFIX_ . 'product_shop` ps ' .
+                'ON (ps.`id_product` = p.`id_product` AND ps.`active` = 1 AND ps.`visibility` IN ("both", "catalog"))'
+            );
         } else {
             $nbProducts = (int) $this->getDatabase()->getValue(
                 'SELECT COUNT(DISTINCT p.`id_product`) ' .
+                'FROM `' . _DB_PREFIX_ . 'product` p ' .
+                'INNER JOIN `' . _DB_PREFIX_ . 'product_shop` ps ON (ps.`id_product` = p.`id_product` AND ps.`active` = 1 AND ps.`visibility` IN ("both", "catalog")) ' .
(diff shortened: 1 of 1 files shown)


Hlavtox:
@Codencode Paolo, I am sorry, but it works fine on my 8.1.7 shop with 4.0.0 module for some reason. 🤔 

And the shop is running since 2015, with holes in the IDs.

Codencode:
@Hlavtox 
Daniel Are you sure the system can process all the products?  
Right now I don't remember exactly, but it might also be that the items get indexed anyway, though it might require more cycles than necessary.

Hlavtox:
@Codencode Yep I think it works fine, but there indeed is an issue.

AI says:
The $cursor vs $nbProducts comparison in indexPrices() is logically flawed because $cursor is a product ID while $nbProducts is just the number of products to index, so the do...while loop often exits after the first batch (especially if product IDs are large).

However, the process still works thanks to the recursive call (or repeated AJAX requests), which continues indexing as long as the cursor advances. The downside is that progress reporting can be inaccurate (it always adds 100 even if fewer rows were processed), and the loop condition is confusing.

A cleaner approach would be to loop while the cursor actually moves forward, or compare against the maximum product ID, so the whole process happens in one loop without relying on recursion.

Codencode:
@Hlavtox,
while reviewing the code I noticed that the problem is that the system performs additional iterations.
I created a copy of the module that sets the conditions to reproduce the issue.

Basically, I set the product table AUTO_INCREMENT to 10000 and create 200 products.
This way the problem can be verified, just:

- install the module
- enter the configuration so that the system creates the additional products (the code is executed by the `getContent()` method)
- run the **Rebuild price index** link by opening it in an empty browser tab.

Here the system prints the number of iterations performed.

I created 2 versions of the module:
- [ps_facetedsearch-test-without-fix.zip](https://github.com/user-attachments/files/21957117/ps_facetedsearch-test-without-fix.zip): the current version of the module with logging instructions and product creation
- [ps_facetedsearch-test-with-fix.zip](https://github.com/user-attachments/files/21957122/ps_facetedsearch-test-with-fix.zip): the version of the module with the fix, also with logging instructions and product creation

Here are the results:
### without fix:
without-fix

### with fix:
with-fix

As you can see, in the first case the system executes the method call 3 times, while in the second case it performs 3 iterations, thus with a single call to the `indexPrices()` method.

PR #1136, opened 2025-04-25
Improvement of sub-categories count

kgleizes:
|              Questions            | Answers
| ----------------------- | -------------------------------------------------------
|        Description?       | Improve products count performance from sub-categories on category page (more details after).
|      Type             | improvement 
|        BC breaks?     | no
|        Deprecations?              | no
|        Fixed ticket?    | No issue opened
|        How to test?     | Display category with several sub-categories and configuration mentionned above.

---------------------
#### Description - Main objectives

When product count and sub-categories is enabled, the SQL query to retrieve the number of products per sub-category on a category page was particularly slow, and could slow down or even bring the MySQL server to its knees. This enhancement splits the process into two queries (one to retrieve the category list) and the second to output the product count. Considerable performance gains (12x to 100x faster depending number of products / categories).

#### More details on configuration used

First of all, we need to display one category with filters and sub-categories count enabled :  
Configuration('_PS_LAYERED_FULL_TREE_') = 1 
Configuration('_PS_LAYERED_SHOW_QTIES_') = 1  
image

Then the category must have one model with sub categories filters enabled (_layered_selection_subcategories_) : 
image

It's better to test without cache of blocks filters enabled and without cache to be abble to see performance before / after. You can add the constant _PS_FACETED_NEWCOUNT_ in your defines.inc.php (or custom) to false to get previous count and compare before / after.

---------------------
#### Explaination of diagnostic with ps profiling tool.

I noticed that the main slow query was pointing the sub-categories count. 
So i tried to go deeper and check which request(s) is | are concerned.
I discover that the sub-categories product count was the main issue here and split the process in two
requests (one big as before and a very small to filter results with concerned categories).

---------------------

#### First case - Shop with 200k+ products - 415 categories - Display of category with 5500 products

**Before**
image

**After**
image

As we can see we have following result : 

- **Before** - **11610 ms** with 253980 results - 2195 queries
- **After** - **931 ms** with 3344771556 results - 2196 queries (**12x faster**)
We have one more query to get the categories concerned by the filters after, and then count is filtered
and we have the same results in output.

---------------------

#### Second case - Shop with 5k+ products - 289 categories - Display of category with 700 products

**Before**
image

**After**
image

As we can see we have following result : 

- **Before** - **3664 ms** with 28896 results - 2009 queries
- **After** - **27.5 ms** with 326163600 results - 2010 queries (**133x faster**)
Same thing one more query to get categories concerned instead of add it in request for count.

---------------------

To be sure that output was the same i had some json exports before after but feel free to make the
same tests and point to me any issue about this and how to reproduce.

There is some conditions remaining to allow us to test ([_PS_FACETED_NEWCOUNT_](https://github.com/202ecommerce/ps_facetedsearch/blob/dev/src/Filters/Block.php#L992)) i can quickly remove
this when all is approved.

**Updated the 2026/06/19** - Tests variables and conditons are now here : 
https://github.com/202ecommerce/ps_facetedsearch/tree/dev-categories-tests

To make easier validation of code there is no more test variable and conditions in this pull request.
If you test this branch, you will see logs files in /var/log/facetedsearch-count-{date}.log

With one GET parameter "newCategoryCount" set to 0 in the category page we will have in the logs
the previous count (time and number of categories).
Without the parameter or set to 1 you will keep this evolution in.
The logs will mention After or Before informations in the results and this allow to be sure that
after and before this update we have the same results.

### DIFF
--- a/src/Adapter/MySQL.php
+++ b/src/Adapter/MySQL.php
@@ -44,6 +44,9 @@ class MySQL extends AbstractAdapter
      */
     const INNER_JOIN = 'INNER JOIN';
 
+    /** @var bool */
+    private $categoryCount = false;
+
     /**
      * {@inheritdoc}
      */
@@ -95,6 +98,12 @@ class MySQL extends AbstractAdapter
         // Prepare mapping for joined tables
         $filterToTableMapping = $this->getFieldMapping();
 
+        if ($this->categoryCount === true) {
+            unset($filterToTableMapping['nleft']);
+            unset($filterToTableMapping['nright']);
+            unset($filterToTableMapping['id_group']);
+        }
+
         // Process and generate all fields for the SQL query below
         $orderField = $this->computeOrderByField($filterToTableMapping);
         $selectFields = $this->computeSelectFields($filterToTableMapping);
@@ -117,6 +126,17 @@ class MySQL extends AbstractAdapter
             foreach ($joinAliasInfos as $tableAlias => $joinInfos) {
                 $query .= ' ' . $joinInfos['joinType'] . ' ' . _DB_PREFIX_ . $joinInfos['tableName'] . ' ' .
                        $tableAlias . ' ON ' . $joinInfos['joinCondition'];
+            }
+        }
+
+        if ($this->categoryCount === true) {
+            foreach($whereConditions as $key => $oneCondition) {
+                if ($oneCondition === "p.id_group='1'") {
+                    unset($whereConditions[$key]);
+                }
+                if (strpos($oneCondition, 'p.nleft') !== false || strpos($oneCondition, 'p.nright') !== false) {
+                    unset($whereConditions[$key]);
+                }
             }
         }
 
@@ -762,8 +782,9 @@ class MySQL extends AbstractAdapter
     /**
      * {@inheritdoc}
      */
-    public function valueCount($fieldName = null)
+    public function valueCount($fieldName = null, $categoryCount = false)
     {
+        $this->categoryCount = $categoryCount;
         $this->resetGroupBy();
         if ($fieldName !== null) {
             $this->addGroupBy($fieldName);
@@ -774,7 +795,6 @@ class MySQL extends AbstractAdapter
         $this->setOrderField('');
 
         $this->copyOperationsFilters();
-
         return $this->execute();
     }
 
@@ -838,6 +858,7 @@ class MySQL extends AbstractAdapter
 
         $operationsFilters = clone $initialPopulation->getOperationsFilters();
         foreach ($operationsFilters as $operationName => $operations) {
+            $message = "Add operation ${operationName} with operation " . json_encode($operations);
             $this->addOperationsFilter(
                 $operationName,
                 $operations
--- a/src/Adapter/MySQL.php
+++ b/src/Adapter/MySQL.php
@@ -131,7 +131,7 @@ class MySQL extends AbstractAdapter
 
         if ($this->categoryCount === true) {
             foreach($whereConditions as $key => $oneCondition) {
-                if ($oneCondition === "p.id_group='1'") {
+                if (strpos($oneCondition, 'p.id_group') !== false) {
                     unset($whereConditions[$key]);
                 }
                 if (strpos($oneCondition, 'p.nleft') !== false || strpos($oneCondition, 'p.nright') !== false) {
@@ -858,7 +858,6 @@ class MySQL extends AbstractAdapter
 
         $operationsFilters = clone $initialPopulation->getOperationsFilters();
         foreach ($operationsFilters as $operationName => $operations) {
-            $message = "Add operation ${operationName} with operation " . json_encode($operations);
             $this->addOperationsFilter(
                 $operationName,
                 $operations
--- a/src/Adapter/MySQL.php
+++ b/src/Adapter/MySQL.php
@@ -825,6 +825,7 @@ class MySQL extends AbstractAdapter
         $this->setOrderField('');
 
         $this->copyOperationsFilters();
+
         return $this->execute();
     }
 
--- a/src/Adapter/MySQL.php
+++ b/src/Adapter/MySQL.php
@@ -102,7 +102,6 @@ class MySQL extends AbstractAdapter
             unset($filterToTableMapping['nleft']);
             unset($filterToTableMapping['nright']);
             unset($filterToTableMapping['id_group']);
-            unset($filterToTableMapping['level_depth']);
         }
 
         // Process and generate all fields for the SQL query below
@@ -135,9 +134,6 @@ class MySQL extends AbstractAdapter
         if ($this->categoryCount === true) {
             foreach ($whereConditions as $key => $oneCondition) {
                 if (strpos($oneCondition, 'p.id_group') !== false) {
-                    unset($whereConditions[$key]);
-                }
-                if (strpos($oneCondition, 'p.level_depth') !== false) {
                     unset($whereConditions[$key]);
                 }
                 if (strpos($oneCondition, 'p.nleft') !== false || strpos($oneCondition, 'p.nright') !== false) {
--- a/src/Adapter/MySQL.php
+++ b/src/Adapter/MySQL.php
@@ -132,9 +132,8 @@ class MySQL extends AbstractAdapter
             }
         }
 
-
         if ($this->categoryCount === true) {
-            foreach($whereConditions as $key => $oneCondition) {
+            foreach ($whereConditions as $key => $oneCondition) {
                 if (strpos($oneCondition, 'p.id_group') !== false) {
                     unset($whereConditions[$key]);
                 }
--- a/src/Adapter/MySQL.php
+++ b/src/Adapter/MySQL.php
@@ -102,6 +102,7 @@ class MySQL extends AbstractAdapter
             unset($filterToTableMapping['nleft']);
             unset($filterToTableMapping['nright']);
             unset($filterToTableMapping['id_group']);
+            unset($filterToTableMapping['level_depth']);
         }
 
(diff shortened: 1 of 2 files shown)


ps-jarvis:
Hello @kgleizes!

sneko:
@Hlavtox do you know which team member could look at this PR? Thank you,

nicohery (approved):

nico-leb (approved):
it would be very usefull for large catalog

Touxten (dismissed):

src/Filters/Block.php
```diff
@@ -989,7 +989,25 @@ private function getCategoriesBlock($filter, $selectedFilters, $idLang, $parent)
...
+        }
+        $results = $filteredSearchAdapter->valueCount('id_category', $categoryCount);
+
+        $categoriesId = [];
+        if ($categoryCount === true) {
+            $query = new \DbQuery();
+            $query->select('id_category');
+            $query->from('category');
+            $query->where('level_depth <= ' . (int) $parent->level_depth + 1);
+            $query->where('nleft > ' . $parent->nleft);
```
jf-viguier:
```suggestion
            $query->where('nleft > ' . (int) $parent->nleft);
```

src/Filters/Block.php
```diff
@@ -989,7 +989,25 @@ private function getCategoriesBlock($filter, $selectedFilters, $idLang, $parent)
...
+            $categoryCount = (bool) _PS_FACETED_NEWCOUNT_;
+        }
+        $results = $filteredSearchAdapter->valueCount('id_category', $categoryCount);
+
+        $categoriesId = [];
+        if ($categoryCount === true) {
+            $query = new \DbQuery();
+            $query->select('id_category');
+            $query->from('category');
+            $query->where('level_depth <= ' . (int) $parent->level_depth + 1);
```
jf-viguier: Should use Configuration::get('PS_LAYERED_FILTER_CATEGORY_DEPTH') instead of 1 ?

src/Adapter/MySQL.php
```diff
@@ -95,6 +98,12 @@ public function getQuery()
         // Prepare mapping for joined tables
         $filterToTableMapping = $this->getFieldMapping();
 
+        if ($this->categoryCount === true) {
+            unset($filterToTableMapping['nleft']);
+            unset($filterToTableMapping['nright']);
+            unset($filterToTableMapping['id_group']);
```
jf-viguier, line 104:
```suggestion
            unset($filterToTableMapping['id_group']);
            unset($filterToTableMapping['level_depth']);
```

src/Adapter/MySQL.php
```diff
@@ -122,6 +131,18 @@ public function getQuery()
             }
         }
 
+
+        if ($this->categoryCount === true) {
+            foreach($whereConditions as $key => $oneCondition) {
+                if (strpos($oneCondition, 'p.id_group') !== false) {
+                    unset($whereConditions[$key]);
+                }
+                if (strpos($oneCondition, 'p.nleft') !== false || strpos($oneCondition, 'p.nright') !== false) {
```
jf-viguier, line 139: add level_depth too ?

src/Filters/Block.php
kgleizes: Thank you, that's updated for nleft and nright below.

You can check and resolve if that's ok for you.

src/Filters/Block.php
kgleizes: Well pointed, i've made again some tests with moving this configuration and that's corrected.
Now we use this configuration to prepare the request.

src/Adapter/MySQL.php
kgleizes: That's updated too, we should see best performances depending the number of categories and product in them.

src/Adapter/MySQL.php
kgleizes: Done.

PrestaEdit (commented):
Thanks for the work and the very detailed benchmarks — the diagnosis is solid and the split (one big global count + a small subtree query filtered in PHP) is a clean idea. The performance gains are real. However, the refactor changes the displayed results in a few real-world cases, which contradicts the "same results before/after" guarantee. A few points to address before this can be merged.

## 🔴 Blocking

### 1. Group access (`id_group` / `category_group`) is dropped entirely
The `id_group` condition (join on `category_group`, `MySQL.php` mapping) restricted the count to categories accessible to the **customer's group**. It is removed from the count query and **never reintroduced** in the new PHP `$categoriesId` query (`Block.php`), which only filters on `nleft/nright/level_depth`.

Result: on shops that restrict categories by group, sub-categories forbidden to the visitor now **reappear in the filter block with their count** — a visibility regression (and a minor leak of restricted categories' existence/volume). The "same results" observation likely holds only because tests were run with the default group / no per-group category restriction.

➡️ Please reintroduce the `category_group` restriction (IN customer groups) in the `$categoriesId` query.

### 2. `active = 1` is no longer guaranteed
The original tree join carried `AND c.active = 1`. The new `$categoriesId` query does not filter `active`.
- With `PS_LAYERED_FILTER_CATEGORY_DEPTH > 0`, the `level_depth` join (with `active=1`) still masks it on the count side.
- But with `depth = 0` (infinite), no `level_depth` condition is added → no category join at all on the count side → **disabled sub-categories can show up**.

➡️ Please add `->where('active = 1')` to the `$categoriesId` query.

## 🟠 Important

### 3. Fragile coupling between the `unset` order and the string-based WHERE stripping
Removing the WHERE conditions relies on `strpos($oneCondition, 'p.nleft')` / `'p.nright'` / `'p.id_group'`. These fields are normally aliased `c.` / `cg.`. This only works because the mapping `unset`s run **before** `computeWhereConditions`, forcing the fallback alias `p`. It's correct today but silently breakable: any reorder/alias change would leave the conditions in place and the query would fail (join removed, column still referenced). Please add an explicit comment about this ordering dependency, or use a less SQL-string-based approach.

### 4. Multistore: `$categoriesId` is not shop-scoped
`->from('category')` has no `category_shop` join. `nleft/nright` are global so the subtree is correct, but the query isn't restricted to the current shop. Worth verifying in multistore.

### 5. No tests
This is a behavior change on counting with no test added, while the module has a test suite. Ideally cover: restricted group (Blocking 1), inactive category (Blocking 2), and `depth = 0` vs `depth > 0`.

## 🟡 Minor
- `PS_LAYERED_FILTER_CATEGORY_DEPTH` is read twice (`addCategoriesBlockFilters` and `getCategoriesBlock`); note `if ($depth)` vs `if ($depth > 0)` — same behavior today, but keep them in sync.
- `in_array($idCategory, $categoriesId)` is O(n) inside the results loop — `array_fill_keys($categoriesId, true)` + `isset()` is cheaper.
- The `$categoriesId` building loop can be replaced by `array_column($resultCategories, 'id_category')`.
- SQL injection: OK, all new params are `(int)` cast. 👍

Overall: great optimization, but Blocking 1 and 2 change what's displayed in legitimate configurations, so they need fixing (with tests) before merge.

kpodemski:
> [!NOTE]

Hi @kgleizes, sorry for the long wait on this one. I rebased your branch on current `dev` and tested it. Two things to report, and one question for you.

## 1. The group access issue @PrestaEdit raised is real

I reproduced it. I removed the Visitor group from a sub-category, then viewed the parent category as a guest:

| build | Categories filter block |
| --- | --- |
| `dev` | Home Accessories (8) |
| this PR | Home Accessories (8), **Stationery (3)** |
| PR + fix | Home Accessories (8) |

A category the visitor is not allowed to see appears in the filter block, with its product count. I confirmed it by logging the generated SQL: after the stripping, the WHERE of the count query is reduced to `c.level_depth<=3`, so the `id_group` condition is gone and nothing puts it back.

The fix is to add the group restriction to your new `$categoriesId` query. I also added `active = 1` and a `category_shop` join for multistore, plus a comment about the ordering between the `unset` calls and the string matching in `computeWhereConditions`.

Two smaller notes on @PrestaEdit's review, after testing:

- The missing `active = 1` does not currently show up, because `getAllCategoriesName()` is called with `$active = true` and the loop skips anything outside that list. Still worth adding, but it is not a live bug.
- Multistore I could not verify, single shop install here.

## 2. I could not reproduce the performance gain

This is the part I would like your help with. I tested on 100 000 products, 400 000 combinations, MariaDB 11.4, with `PS_LAYERED_FULL_TREE=1`, `PS_LAYERED_SHOW_QTIES=1` and the block cache off. I grew the category tree three times to get closer to your shop:

| categories | sub-categories under the browsed one | count query, `dev` | count query, PR |
| --- | --- | --- | --- |
| 34 | 0 | 0.545s | 0.541s |
| 418 | 16 | 0.162s | 0.163s |
| 2 434 | 100 | 0.166s | 0.167s |

Page time did not move either, against controls that stayed flat. I also tried `PS_LAYERED_FILTER_CATEGORY_DEPTH = 0`, same result.

Here the sub-category count is about 0.17s of a 4.4s page, so even removing it entirely could not produce 12x. Your "before" was 11.6s for that one query, which is 70 times slower than what I measure on a much larger tree. That gap looks like a query plan difference rather than a data size difference, so I am fairly sure your shop hits something this environment does not.

## Could you share a bit about your setup?

- MySQL or MariaDB, and which version?
- `EXPLAIN` of the slow count query on your shop, before your change
- Roughly how many products, and how deep the category tree is
- Value of `PS_LAYERED_FILTER_CATEGORY_DEPTH`

With the `EXPLAIN` we can find what the optimizer does differently, and then decide whether the split is the right fix or whether an index or a smaller change solves it.

To be clear, I am not saying your numbers are wrong. I just cannot reproduce them yet, and I would rather understand why before merging a change to this query.

kpodemski (changes requested):
See my comment above.

PR #1156, opened 2025-07-09
Provide proper product order on search page

Hlavtox:
| Questions         | Answers
| Description?      | Fixes search result ordering in case of ordering by relevance. ✅ It provides the same results as the default page. ✅ I made sure that it returns `Relevance` sort order with proper asc/desc for search and other pages. ✅ It properly sorts with relevance orders if passed to the URL. ✅ Both default relevance and manually selected in URL by `&order=product.position.desc` return the same results.
| Type?             | bug fix
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | 
| How to test?      | Result order should be the same with and without faceted search installed.
| Sponsor company   | 

Without faceted search
withouzt

With faceted search
with

### DIFF
--- a/ps_facetedsearch.php
+++ b/ps_facetedsearch.php
@@ -1748,7 +1748,7 @@ VALUES(' . $last_id . ', ' . (int) $idShop . ')');
                 'cacheable' => false,
             ],
             'search' => [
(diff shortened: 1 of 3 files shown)


Hlavtox:
@sneko Just finished the PR, waiting for tests to finish, then it will be ready. :-)

Works, tested, both directions, all cases, everything.

SiraDIOP (approved):

PR #1163, opened 2025-08-09
Fix category facet so it matches deep subcategories

### LINKED ISSUE #29727: Facets search by second level subcategory does not work
### Prerequisites

- [X] I understand and accept the project's [code of conduct](https://github.com/PrestaShop/PrestaShop/blob/develop/CODE_OF_CONDUCT.md)

### Describe the bug and add screenshots

Hi!
I am trying to make a search by second level subcategory but does not seem to work. What I mean for second level subcategory is the following:

Clothes
-> Men (First level subcategory, pending from Clothes)
----> Child (Second level subcategory, pending from Men)

### Expected behavior

That when I filter on facets by second level subcategory the filter would be added and work properly. Now it seems that it is ignoring second level subcategory filters because they are not being applied right now.

### Steps to reproduce

- Installing a cleaning Prestashop v1.7.8.7 with demo data
- By default, it gives to you only 1 level of subcategory for example Clothes > Men so I have created a new pending category for Men called "Child" and I have added some products to it, similar to the previous schema.
- I have configured "Category filter depth" to 0 on "Faceted Search" module to do not have limit on category filter and added the new category on the template.
- Also indexed all mixing prices, rebuild entire price index, build attributes and features indexes and clear caché from "Faceted Search" module.

After made the configuration I go to the front:
If I go to Clothes section, it let me filter by "Men" but not by "Child". Only lets me filter by first level subcategory, like "Men" or "Women", but if I go to "Men" section, it lets me filter by "Child", because right now "Child" its first level subcategory from "Men", so the problem is only when I try to filter by second level subcategory. For example: when I am on "Clothes" and I try to filter by "Child".

1.7.8.7

7.4

ps_facetedsearch v3.8.0

Pianczetti:
| Questions         | Answers
| Description?      | ps_facetedsearch: fix category facet so it matches deep subcategories (not only direct children). Implemented entirely in module to keep backward compatibility with PS 1.7/8/9. Uses a single subtree-scoped query (nleft/nright) to resolve category names within the current category subtree. Adds a batch multi-name lookup to minimize DB round-trips and a safe BFS fallback when needed.
| Type?             | bug fix
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | Fixes [PrestaShop/Prestashop#29727](https://github.com/PrestaShop/PrestaShop/issues/29727).
| How to test?      | 1) Enable ps_facetedsearch and the “Categories” facet. <br/>2) Create a tree: Parent (P) → Child (C) → Grandchild (G). Assign products to all categories <br/>3) Rebuild the module’s index and clear cache. <br/>4) Go to FO category page of P. <br/>5) In the Categories facet, select G. Expected: product list shows only products from G (previously deeper subcategories were not matched).
| Sponsor company   |

### DIFF
--- a/src/Filters/Converter.php
+++ b/src/Filters/Converter.php
@@ -24,6 +24,7 @@ use Category;
 use Configuration;
 use Context;
 use Db;
+use Shop;
 use Manufacturer;
 use PrestaShop\Module\FacetedSearch\Definition\Availability;
 use PrestaShop\Module\FacetedSearch\Filters;
@@ -448,15 +449,28 @@ class Converter
                     break;
                 case self::TYPE_CATEGORY:
                     if (isset($receivedFilters[$filterLabel])) {
+                        // Collect valid names
+                        $requestedNames = [];
                         foreach ($receivedFilters[$filterLabel] as $queryFilter) {
-                            /*
-                             * This works only for categories that are child of the category we are browsing (or home category).
-                             * Categories deeper in the tree will never be found. This could be fixed by providing a unique ID
-                             * to the URL.
-                             */
-                            $categories = Category::searchByNameAndParentCategoryId($idLang, $queryFilter, (int) $idCategory);
-                            if ($categories) {
-                                $searchFilters[$filter['type']][] = $categories['id_category'];
+                            if (is_string($queryFilter) && $queryFilter !== '') {
+                                $requestedNames[] = $queryFilter;
+                            }
+                        }
+
+                        if (!empty($requestedNames)) {
+                            // Batch lookup by subtree; returns name => id_category mapping
+                            $foundMap = $this->findCategoriesByNamesInSubtree($idLang, $requestedNames, (int) $idCategory);
+
+                            foreach ($requestedNames as $queryFilter) {
+                                if (isset($foundMap[$queryFilter])) {
+                                    $searchFilters[$filter['type']][] = (int) $foundMap[$queryFilter];
+                                    continue;
+                                }
+                                // Fallback per-name (SQL single + BFS) if not resolved in batch
+                                $categoryRow = $this->findCategoryByNameInSubtree($idLang, $queryFilter, (int) $idCategory);
+                                if ($categoryRow) {
+                                    $searchFilters[$filter['type']][] = (int) $categoryRow['id_category'];
+                                }
                             }
                         }
                     }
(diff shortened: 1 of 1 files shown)


ps-jarvis:
Hello @Pianczetti!

AmandaIsFive:
Amazing, thank you!! Apparently it works perfectly.  I hope they approve your PR soon to improve the module's functionality.

kpodemski:
> [!NOTE]

Hi @Pianczetti, thanks for this, and sorry for the wait. I confirmed the bug, kept your fix, and pushed a simplification to your branch. Details below, and I would like you to validate it.

## The bug is real and still present

It only appears when "Category filter depth" is not 1. At the default of 1 only direct children are offered, so nothing looks broken. Set it to 0 and the facet offers grandchildren that do nothing when clicked.

On a small shop, browsing Home which has 19 products:

| clicked in the facet | `dev` | your branch |
| --- | --- | --- |
| Clothes (direct child) | 2 | 2 |
| Men (grandchild) | **19** | 1 |
| Women (grandchild) | **19** | 1 |
| Stationery (grandchild) | **19** | 3 |

It is worse on a large catalogue. On a 100 000 product category, filtering by a level 4 sub-category returns **the whole 100 000** on `dev`, and 29 with the fix. So the customer clicks a filter and gets the heaviest possible page instead of a filtered one.

## What I changed on your branch

> [!IMPORTANT]
> Your branch was rebased and force updated, so its history changed. Before you push again:
> ```
> git fetch origin && git reset --hard origin/fix/facets-search-subcategory-29727
> ```
> Otherwise your next push will conflict. Sorry, I should have said this in the first version of this comment.

There are now three commits: yours, my change below, and a small one moving the `use Shop;` import so PHP-CS-Fixer passes. It was ordered after `Db`, and this project sorts every import alphabetically, including the namespaced ones, so it belongs at the end.

I rebased it on current `dev` and replaced the three step lookup with a single query. The reason is performance.

`findCategoryByNameInSubtreeFallback()` walks the whole subtree, two queries per category, whenever a name does not resolve. That happens for any renamed category, any old bookmark and any crawler hitting a stale URL. On a shop with 2 434 categories:

| build | page with 1 unknown category name | SQL statements |
| --- | --- | --- |
| `dev` | 2.650s | 511 |
| your branch | 3.109s | about 4 900 extra |
| after the change | 2.512s | 505 |

The single name SQL lookup could go too. It runs the same query as the batch one, `name = 'x'` instead of `name IN (...)`, with the same subtree, active and home filters. If the batch found nothing, that one cannot find anything either.

So the resolution is now one query, and it is slightly cheaper than `dev` because a single `IN` replaces one query per name.

Two smaller fixes in the same commit:

- Added `ORDER BY level_depth ASC, id_category ASC`. The batch query had no order while the single one ordered by depth, so two same named categories could resolve differently depending on which path ran.
- Changed the `category_shop` join to `INNER JOIN`. The comment said it was equivalent to `Shop::addSqlAssociation`, but a `LEFT JOIN` with nothing in the `WHERE` does not filter anything, so in multishop a category from another shop could match.

Your `c.active = 1` is a real improvement by the way. `Category::searchByNameAndParentCategoryId()` in core does not have it, so disabled categories used to resolve.

Result: 79 lines instead of 167, 116 tests pass, and the deep sub-category fix behaves exactly as before.

## Could you validate?

Please check the branch on your own setup, especially any case that made you add the tree walk in the first place. If you know a shop where the SQL lookup fails and the traversal saves it, I want to hear about it, because that would mean the single query needs fixing rather than the fallback being removed.

## Known limitation, not for this PR

Category names are not unique in a tree. With two categories named "Men" in different branches, the facet lists two "Men" entries with different counts, and both produce the same URL, so one of them is unreachable. My ordering makes the choice stable, not correct. Fixing it properly needs an id in the URL, which is what the original comment in the code already said. We will handle that separately, so please do not treat it as blocking here.

Hlavtox:
@kpodemski @Pianczetti I think a proper permanent solution would be to also add the category ID to the URL. We can then remove this searching by name. WDYT.

kpodemski:
@Hlavtox eventually it's the only way to go, but if one uses a module's URL somewhere then it's quite a big BC Break

Hlavtox:
@kpodemski Yep, but we could bump the version to v6?

I experienced with it here - https://github.com/PrestaShop/ps_facetedsearch/pull/779, it wasn't that difficult.

kpodemski:
@Hlavtox I think only after finalizing what's left, there a lot of good things coming to 5.1, better to finish it

PR #1170, opened 2025-08-23
feat: add hooks to extend faceted search filters and supported contro…

Codencode:
| Questions         | Answers
| Description?      | This PR aims to make the module more flexible by allowing the use of filters in ad hoc controllers developed by other modules. Two new hooks have been added: `actionFacetedSearchSetSupportedControllers`: allows adding a new controller to display it in the filter template configuration, `actionFacetedSearchFilters`: allows setting the starting filter on which "to apply the filters."
| Type?             | new feature
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     |
| How to test?      |1. Install the module [cnc_sampleproductlisting.zip](https://github.com/user-attachments/files/21950718/cnc_sampleproductlisting.zip) - 2. - Edit template filter (in configuration of Faceted search module) and select new page controller **Product listing filters** - 3. go to module configuration of cnc_sampleproductlisting module - 4. Copy the URL Link listing controller and open it in the browser - 5.View product with id 1,2,3,4 which are the products that the module sets in the display using the "hookActionFacetedSearchFilters" hook. - 6. Interact with the filters to verify that the system applies the filters only to the 4 products.
| Sponsor company   | @Codencode 

### This is the template configuration page
The new controller created by the sample module appears here.
edit-tilter-template

### This is the page of the front controller created by the sample module.

front-controller

### DIFF
--- a/ps_facetedsearch.php
+++ b/ps_facetedsearch.php
@@ -1726,7 +1726,7 @@ VALUES(' . $last_id . ', ' . (int) $idShop . ')');
 
     public function initializeSupportedControllers()
     {
-        $this->setSupportedControllers([
+        $supportedControllers = [
             'category' => [
                 'name' => $this->trans('Category', [], 'Modules.Facetedsearch.Admin'),
                 'cacheable' => true,
@@ -1755,6 +1755,23 @@ VALUES(' . $last_id . ', ' . (int) $idShop . ')');
                 'name' => $this->trans('Search', [], 'Modules.Facetedsearch.Admin'),
                 'cacheable' => false,
             ],
-        ]);
+        ];
+
+        $additionalSupportedControllers = Hook::exec(
+            'actionFacetedSearchSetSupportedControllers',
+            [],
+            null,
+            true,
+            true,
+            false,
+            null,
+            true
+        );
+
+        if ($additionalSupportedControllers && is_array($additionalSupportedControllers)) {
+            $supportedControllers = array_merge($supportedControllers, $additionalSupportedControllers);
+        }
+
+        $this->setSupportedControllers($supportedControllers);
     }
 }
--- a/ps_facetedsearch.php
+++ b/ps_facetedsearch.php
@@ -1757,20 +1757,12 @@ VALUES(' . $last_id . ', ' . (int) $idShop . ')');
             ],
         ];
 
-        $additionalSupportedControllers = Hook::exec(
+        Hook::exec(
             'actionFacetedSearchSetSupportedControllers',
-            [],
-            null,
-            true,
-            true,
-            false,
-            null,
-            true
+            [
+                'supportedControllers' => &$supportedControllers,
+            ]
         );
-
-        if ($additionalSupportedControllers && is_array($additionalSupportedControllers)) {
-            $supportedControllers = array_merge($supportedControllers, $additionalSupportedControllers);
-        }
 
         $this->setSupportedControllers($supportedControllers);
     }
--- a/src/Product/Search.php
+++ b/src/Product/Search.php
@@ -25,6 +25,7 @@ use Configuration;
 use Context;
 use FrontController;
 use Group;
+use Hook;
 use PrestaShop\Module\FacetedSearch\Adapter\AbstractAdapter;
 use PrestaShop\Module\FacetedSearch\Adapter\MySQL as MySQLAdapter;
 use PrestaShop\Module\FacetedSearch\Definition\Availability;
@@ -450,6 +451,14 @@ class Search
                 empty($productPool) ? ['NULL'] : $productPool
             );
         }
+
+        Hook::exec(
+            'actionFacetedSearchFilters',
+            [
+                'search' => $this,
+                'query' => $this->query,
+            ]
+        );
     }
 
     /**
--- a/tests/php/FacetedSearch/MockProxy.php
+++ b/tests/php/FacetedSearch/MockProxy.php
@@ -144,3 +144,9 @@ class FrontController extends MockProxy
     // Redeclare to use this instead MockProxy::mock
     protected static $mock;
 }
+
+class Hook extends MockProxy
+{
+    // Redeclare to use this instead of MockProxy::$mock
+    protected static $mock;
+}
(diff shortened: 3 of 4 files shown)


Codencode:
@Hlavtox, Daniel, what do you think?
In my opinion, it could be useful to have the possibility to add filters in an "ad hoc" controller.
I added 2 new hooks, I don't know if they need to be created "explicitly" and therefore added to the database during the module installation.

Codencode:
If the change is okay, I think I need to check the tests, but I'll wait for some feedback on the idea first.
Furthermore, I also think I need to review the execution of the `actionFacetedSearchSetSupportedControllers `hook for PS 1.7.6 versions, since the `$chain` parameter was added in PS 1.7.7.x.

ps_facetedsearch.php
```diff
@@ -1755,6 +1755,23 @@ public function initializeSupportedControllers()
                 'name' => $this->trans('Search', [], 'Modules.Facetedsearch.Admin'),
                 'cacheable' => false,
             ],
-        ]);
+        ];
+
+        $additionalSupportedControllers = Hook::exec(
```
Hlavtox: What if we just pass the controllers array as reference? It will work easier and no need for chaining.

Hlavtox (commented):
Just a small comment, but love this. I have “on sale products” and “our tips” controllers on my shop and I need to keep modified version of faceted search.

This could maybe also allow implementing “hard filters” like car models easier.

ps_facetedsearch.php
Codencode: Good idea, I think it would work just as well and, as you already mentioned, it would solve the compatibility issue with PS 1.7.6!

Codencode:
> Just a small comment, but love this. I have “on sale products” and “our tips” controllers on my shop and I need to keep modified version of faceted search.
> 
> This could maybe also allow implementing “hard filters” like car models easier.

So, is the idea a good one?

I've never used this module on my clients' e-commerce sites because I had a lot of problems with the old module (I'm talking about the one from PS version 1.6). Because of that, I've always used an external module that I adapted to my needs.

Codencode:
@Hlavtox, I made the change; the parameter is now passed by reference.

Hlavtox (dismissed):
Love it, good job

Hlavtox:
Cc @kpodemski can we merge? It's just a hook. :-)

CI failure is not related

kpodemski:
@Hlavtox yep

Hlavtox:
@kpodemski Do you think of any other needed extension points in the module? Or is it OK like this?

Codencode:
@Hlavtox @kpodemski  
If you can, please take a look at this PR as well: https://github.com/PrestaShop/ps_facetedsearch/pull/1124

Codencode:
Hi @Hlavtox,
do you think we can do something about this error?
https://github.com/PrestaShop/ps_facetedsearch/actions/runs/17178573482/job/48739226195?pr=1170

Codencode:
ping for review @PrestaShop/committers

ga-devfront (dismissed):
Really nice job thank you for this contribution !
About PHPUnit test I'm going to investigate on it if I got the time.

Codencode:
@ga-devfront can this PR be tested? or would you prefer to wait?

ga-devfront:
> @ga-devfront can this PR be tested? or would you prefer to wait?

I think you just have to moke `Hook` inside `https://github.com/PrestaShop/ps_facetedsearch/blob/dev/tests/php/FacetedSearch/Product/SearchTest.php` file and that's all. But QA can pass on your PR.

Codencode:
@ga-devfront I've made the changes, but I'm not sure I fully understood your suggestion and I don't know if the change is actually correct, especially since there's still a PHPStan error.

Can you check?

Thank you very much 🙏

ga-devfront (approved):
Everything is fine, it's normal for phpstan to be red in the latest version, this is the case for all current Core repositories.

Hlavtox (approved):

Hlavtox:
@kpodemski @Robin-Fischer-PS Can we merge guys?

ga-devfront:
@Hlavtox waiting for QA :heart:

SiraDIOP:
Hello @Codencode, 

Thanks for your PR, I tested it and it's good for me 

https://github.com/user-attachments/assets/1c9ba308-4091-4fc3-8a4f-b52b3820ca8f

SiraDIOP (approved):

ps-jarvis:
PR merged, well done!

Message to @PrestaShop/committers: do not forget to milestone it before the merge.

mathieuhen:
Finally! A hook to customise the query, great! We've been waiting for this for a long time!
I just tested it and it works well.
Unless the module's cache system is enabled.
I think you should also add another hook in the function that generates the key for the cache hash, like this, for example.

```
private function generateCacheKeyForQuery(ProductSearchQuery $query, array $facetedSearchFilters) {

        $context = $this->module->getContext();

        $filterKey = $query->getQueryType();
        if ($query->getQueryType() == 'category') {
            $filterKey .= $query->getIdCategory();
        } elseif ($query->getQueryType() == 'manufacturer') {
            $filterKey .= $query->getIdManufacturer();
        } elseif ($query->getQueryType() == 'supplier') {
            $filterKey .= $query->getIdSupplier();
        }

        \Hook::exec(
            'actionFacetedSearchCacheKeyGeneration',
            [
                'filterKey' => &$filterKey,
                'query' => $query,
                'facetedSearchFilters' => &$facetedSearchFilters,
            ]
        ); 
       

        $filterHash = md5(
            sprintf(
                '%d-%d-%d-%s-%d-%s',
                (int) $context->shop->id,
                (int) $context->currency->id,
                (int) $context->language->id,
                $filterKey,
                (int) $context->country->id,
                serialize($facetedSearchFilters)
            )
        );

        return $filterHash;
    }
```

What do you think?

Codencode:
Hello @mathieuhen, thank you for the suggestion about the new Hook, indeed I hadn't thought about caching.
I created this PR: https://github.com/PrestaShop/ps_facetedsearch/pull/1195 which adds it, what do you think?

Thank you

mathieuhen:
That sounds perfect :)

PR #1189, opened 2025-11-17
Remove condition to get currency

Hlavtox:
| Questions         | Answers
| Description?      | New take on https://github.com/PrestaShop/ps_facetedsearch/pull/784 - to remove code that is used only in tests. We always have this method available now. When getting price specifications, there was a condition that checkes if price specs can be retrieved in the "new way", since 1.7.6, `if (isset($this->context->currentLocale) && method_exists($this->context->currentLocale, 'getPriceSpecification')) {`, or if it should return some hardcoded fallback specifications. Because the minimum supported version is 1.7.6, this condition is no longer needed and the hardcoded price specs below were used only in tests.
| Type?             | refacto
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | 
| How to test?      | Just install it and see that price filter works fine in FO. :-) 

### DIFF
--- a/src/Filters/Block.php
+++ b/src/Filters/Block.php
@@ -1030,71 +1030,29 @@ class Block
      */
     private function preparePriceSpecifications()
     {
-        /* @var Currency */
-        $currency = $this->context->currency;
-        // New method since PS 1.7.6
-        if (isset($this->context->currentLocale) && method_exists($this->context->currentLocale, 'getPriceSpecification')) {
-            /* @var PriceSpecification */
-            $priceSpecification = $this->context->currentLocale->getPriceSpecification($currency->iso_code);
-            /* @var NumberSymbolList */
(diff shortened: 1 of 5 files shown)


Codencode (approved):

kpodemski:
@Hlavtox 

Could you fill "How to test"? Maybe checking filtering in different currencies or something like that?

Hlavtox:
@kpodemski Well, if you look on the code, the part I deleted was never called, because we have minimal compatibility from 1.7.7 and the code is there since 1.7.5 or 1.7.6.

SiraDIOP:
Hello @Hlavtox, 

Thanks for your PR, i tested it and it's good for me the filter price works correctly

SiraDIOP (approved):

PR #1226, opened 2026-05-04
product price indexation omit country

Markus-Gurkcity:
| Questions         | Answers
| Description?      | Indexing product prices creates a large number of entries in the database, making the table very large. In many cases, however, some data can be omitted, for example, if you only want to include one country in the index.<br>With this fix, instead of writing prices for every country, only the price for the shop's default country is written to the index table.
| Type?             | improvement / new feature
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | 
| How to test?      | 
| Sponsor company   | 

This is a example for 14.000 products:

Screenshot 2026-05-04 124629
Screenshot 2026-05-04 124713

And here after the fix:
Screenshot 2026-05-04 132008

### DIFF
--- a/ps_facetedsearch.php
+++ b/ps_facetedsearch.php
@@ -210,7 +210,7 @@ class Ps_Facetedsearch extends Module implements WidgetInterface
             Configuration::updateValue('PS_LAYERED_FILTER_BY_DEFAULT_CATEGORY', 0);
             Configuration::updateValue('PS_USE_JQUERY_UI_SLIDER', 1);
             Configuration::updateValue('PS_LAYERED_DEFAULT_CATEGORY_TEMPLATE', 0);
-            Configuration::updateValue('PS_LAYERED_OMIT_COUNTRIES', 0);
+            Configuration::updateValue('PS_LAYERED_INDEX_COUNTRIES', 1);
 
             $this->psLayeredFullTree = 1;
 
@@ -401,7 +401,7 @@ class Ps_Facetedsearch extends Module implements WidgetInterface
             }
         }
 
-        $omit_countries = (bool) Configuration::get('PS_LAYERED_OMIT_COUNTRIES');
+        $index_countries = (bool) Configuration::get('PS_LAYERED_INDEX_COUNTRIES');
 
         $shopList = Shop::getShops(false, null, true);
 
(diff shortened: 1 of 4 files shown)


ps-jarvis:
Hello @Markus-Gurkcity!

kpodemski (dismissed):

Hlavtox (changes requested):
Good, but I would personally prefer to reverse the condition - instead of “don't index countries” do “index countries”.

Also, what about storing 0 as id_country instead? It would be probably cleaner.

ps-jarvis:

    - [ ] `Index prices by countries`
    - [ ] `Index prices by countries option allows you to enable or disable the indexing of prices for different countries in the layered navigation filters. Turn this off when you do not need country-specific pricing.`

Markus-Gurkcity:
@Hlavtox I reversed the condition from omit countries to index countries. And I changed the default country to 0 as requested

kpodemski:
> [!NOTE]

Thanks for this — trimming the price index is a genuinely useful idea, and the numbers in your description make the case well. I rebased the two commits onto current `dev` and pushed the result to your branch, so the conflicts are gone (they were in `ps_facetedsearch.php` and `src/Adapter/MySQL.php`, both just your code landing next to newer code — your logic is unchanged). Then I tested it on a 9.2 shop. It needs rework before it can go in.

### The write side and the read side disagree, so the default install breaks price filtering

With `PS_LAYERED_INDEX_COUNTRIES = 1`, which is what `install()` and `upgrade-4.0.4.php` set:

- the indexer writes only the default country — `$countries = [['id_country' => PS_COUNTRY_DEFAULT]]`
- the reader joins on `psi.id_country = 0` — `$countryCondition = ' AND psi.id_country = 0'`

Nothing has `id_country = 0`, so the join matches no rows and the price filter disappears completely. Indexing one product on my shop wrote a single row with `id_country = 8` while the query looked for `0`.

`dev`, same category:

dev-price-facet-present

This branch, same category — `Price` is gone from *Filter By*:

pr-price-facet-gone

### The flag means opposite things in different places

| | `= 1` ("Yes") | `= 0` ("No") |
| --- | --- | --- |
| tax rate query | all countries | `AND c.id_country = 0` → matches nothing |
| `$shopCountries` fallback | all shop countries | synthetic `id_country = 0` |
| index loop | **only the default country** | **all countries** |
| reader join | `psi.id_country = 0` | `psi.id_country = PS_COUNTRY_DEFAULT` |

The first two rows read "1 = per country", the last two read "1 = single country". Setting it to `0` — the option labelled *No* — is what makes the indexer write a row per country, which is the opposite of the feature's purpose. It also writes a stray `id_country = 0` row.

### The reader no longer uses the visitor's country

`dev` joins on `psi.id_country = $this->getContext()->country->id`. Both branches here use a constant instead, so once several countries are indexed a visitor from a non-default country is matched against the default country's prices. Whatever the flag ends up meaning, the "index everything" path needs to keep reading the context country.

### Smaller items

- The switch cannot be saved: the template posts `ps_layered_filter_index_countries`, `getContent()` reads `Tools::getValue('ps_layered_index_countries')`. The value is always empty, so saving the settings form writes `0` regardless of what you clicked.
- `upgrade-4.0.4.php` is the wrong place for the new default — the module is at `5.0.0`, so shops already past 4.0.4 never run it and end up with the key unset (`false`), which is the "index every country" branch. A fresh install gets `1`, an upgrade gets `0`: the two disagree.
- PHPUnit is red (`Undefined index: PS_LAYERED_INDEX_COUNTRIES`, 46 errors) because `getFieldMapping()` now calls `Configuration::get()` and the test doubles don't know the key. This is on the PR's own CI too, not something the rebase introduced.
- `src/Adapter/MySQL.php` also carries unrelated trailing-whitespace changes in the `specific_price` join, which will keep causing conflicts — worth dropping.

### Suggestion

Pick one direction for the flag and use it consistently — reading it as "index every country" (`1` = today's behaviour) is the least surprising, keeps the default identical to `dev`, and makes the reader `id_country = context->country->id` when on and `id_country = PS_COUNTRY_DEFAULT` when off. Then the indexer just needs the country list narrowed to the default country when the flag is off, and nothing else has to change.

Note your fork's `dev` was force-updated by that rebase, so pull with `git fetch origin && git reset --hard origin/dev` before you continue, or your next push will conflict.

Hlavtox:
@kpodemski @Markus-Gurkcity I have one more idea. What about indexing countries if `SELECT 1 FROM specific_price WHERE id_country > 0 LIMIT 1`?

Hlavtox:
@kpodemski @Markus-Gurkcity Next day, maybe we could have an even better solution. 😄 And we could even improve the price filter for group-specific prices at the same time.

**For each product, depending on whether it has specific pricing or not:**
* If the product has no country-specific pricing, index it once with `id_country = 0`.
* If the product has any country-specific pricing, index it for all relevant countries.
* If the product has no group-specific pricing, index it once with `id_group = 0`.
* If the product has any group-specific pricing, index it for all relevant groups.

**This also means we don't necessarily need the full `country × group` matrix for every product:**
* no country/group-specific pricing → `0 / 0`
* country-specific only → `X / 0`
* group-specific only → `0 / X`
* both → `X / X`

**We either index `0` or the individual values, never both. Then the query can simply match:**
`id_country IN (0, current_country)`
`id_group IN (0, current_group)`

This would make the index automatically **as small as possible** without any configuration switch, while still keeping country/group-specific prices correct.

PR #1241, opened 2026-06-19
Align module on PrestaShop 8.2+: add 9.0.3 PHPStan job, fix errors and bump to 5.0.0

mattgoud:
| Questions         | Answers
| ------------------| -------------------------------------------------------
| Description?      | Align the module on PrestaShop 8.2+: add a 9.0.3 PHPStan job, fix PHPStan errors instead of suppressing them, bump the module to 5.0.0
| BC breaks?        | yes
| Deprecations?     | no
| Fixed ticket?     | Part of PrestaShop/PrestaShop#41648
| Sponsor company   | ~
| How to test?      | The "PHP tests" workflow must stay green, including the new PHPStan jobs for PrestaShop 9.0.3 (PHP 8.1 & 8.4). None of the neon configs contain any `ignoreErrors` anymore.

### 1. Add the missing PrestaShop 9.0.x PHPStan coverage

The PHPStan matrix only covered `8.2.x`, `9.1.x` and `develop` — there was no job for **PrestaShop 9.0.x**, unlike the other modules aligned under Epic #41648. This adds a `9.0.3` job (PHP 8.1 & 8.4 — 8.5 is unsupported on 9.0.x) by switching the `phpstan` job matrix to an explicit `include` list, plus its `tests/php/phpstan/phpstan-9.0.3.neon` config.

### 2. Fix the PHPStan errors instead of suppressing them

Every `ignoreErrors` entry is removed from **all** neon configs (`8.2.x`, `9.0.3`, `9.1.x` and `develop`); the underlying issues are fixed in code:

- **`Currency::getCurrencies()`** (`ps_facetedsearch.php`) — a `new Shop($idShop)` object was passed as the 3rd `bool $groupBy` parameter. `getCurrencies()` already scopes on the current shop, so the bogus argument is dropped (no behaviour change — the object only forced `groupBy` on, a no-op once currencies are scoped to a single shop).
- **`Tools::displayDate()`** (`ps_facetedsearch.php`) — removed the dead `else` branch using the legacy 3-argument signature; it targeted PrestaShop < 8.0 while the module now requires `>= 8.2`, so it was unreachable.
- **`LegacyControllerContext` calls** (`ps_facetedsearch.php`, `src/Hook/ProductSearch.php`, `src/Hook/AbstractHook.php` + the 4 admin form hooks) — on PrestaShop 9.x `Context::$controller` is typed as the PHPStan-opaque `LegacyControllerContext`, hiding legacy controller members (`addJS`/`addCSS`/`addJqueryUi`/`registerStylesheet`/`registerJavascript` and the `$default_form_language` property). They are now accessed through a concretely-typed `FrontController`/`AdminController` (local `@var` for the front, a small `getAdminController()` helper in `AbstractHook` for the back office).
- **Dead strict comparison** (`src/Product/SearchProvider.php`) — in `updateQueryString()`, `$params` is filled by `parse_str()` (never null) and the preceding loop already strips null values, so the `null === $param` test was dead code and is removed.

### 3. Align minimum compatibility on PrestaShop 8.2+ and bump the version

`ps_versions_compliancy` minimum is raised from `8.1.0` to **`8.2.0`**, matching what the CI actually verifies (the shared PHPStan action targets 8.2+). The module version is bumped accordingly to **5.0.0** (major, since dropping 8.1 support is a BC break). `BC breaks: yes` for that reason.

### 4. Minor cleanup

Dropped a few unused `foreach` value/index variables flagged by static analysis (iterating with `array_keys()` where only the key is used).

### DIFF
--- a/ps_facetedsearch.php
+++ b/ps_facetedsearch.php
@@ -618,7 +618,7 @@ class Ps_Facetedsearch extends Module implements WidgetInterface
 
                 // Associate shops in case of multistore
                 if (isset($_POST['checkBoxShopAsso_layered_filter'])) {
-                    foreach ($_POST['checkBoxShopAsso_layered_filter'] as $idShop => $row) {
+                    foreach (array_keys($_POST['checkBoxShopAsso_layered_filter']) as $idShop) {
                         $filterValues['shop_list'][] = (int) $idShop;
                     }
                 } else {
@@ -638,7 +638,7 @@ class Ps_Facetedsearch extends Module implements WidgetInterface
                 }
 
                 // Add filters themselves
-                foreach ($_POST as $key => $value) {
+                foreach (array_keys($_POST) as $key) {
                     if (!preg_match('~^(?P<key>layered_selection_.*)(?<!_filter_)(?<!type)(?<!show_limit)$~', $key, $matches)) {
                         continue;
                     }
@@ -1051,7 +1051,7 @@ class Ps_Facetedsearch extends Module implements WidgetInterface
         ];
 
         // Add all stable controllers (except search)
-        foreach ($this->getSupportedControllers() as $controller_name => $data) {
+        foreach (array_keys($this->getSupportedControllers()) as $controller_name) {
             if ($controller_name != 'search') {
                 $filterData['controllers'][] = $controller_name;
             }
@@ -1146,7 +1146,7 @@ class Ps_Facetedsearch extends Module implements WidgetInterface
 
             $filterData['shop_list'] = $shopList;
 
-            foreach ($c as $idCategory => $category) {
+            foreach (array_keys($c) as $idCategory) {
                 if (!in_array($idCategory, $filterData['categories'])) {
                     $filterData['categories'][] = $idCategory;
                 }
@@ -1185,7 +1185,7 @@ class Ps_Facetedsearch extends Module implements WidgetInterface
 
                 // Attribute filter
                 if (is_array($attributeGroupsById) && count($attributeGroupsById) > 0) {
-                    foreach ($a as $kAttribute => $attribute) {
+                    foreach (array_keys($a) as $kAttribute) {
                         if (!isset($doneCategories[(int) $idCategory]['a' . (int) $attributeGroupsById[(int) $kAttribute]])) {
                             $filterData['layered_selection_ag_' . (int) $attributeGroupsById[(int) $kAttribute]] = ['filter_type' => Converter::WIDGET_TYPE_CHECKBOX, 'filter_show_limit' => 0];
                             $doneCategories[(int) $idCategory]['a' . (int) $attributeGroupsById[(int) $kAttribute]] = true;
@@ -1196,7 +1196,7 @@ class Ps_Facetedsearch extends Module implements WidgetInterface
 
                 // Features filter
                 if (is_array($featuresById) && count($featuresById) > 0) {
-                    foreach ($f as $kFeature => $feature) {
+                    foreach (array_keys($f) as $kFeature) {
                         if (!isset($doneCategories[(int) $idCategory]['f' . (int) $featuresById[(int) $kFeature]])) {
                             $filterData['layered_selection_feat_' . (int) $featuresById[(int) $kFeature]] = ['filter_type' => Converter::WIDGET_TYPE_CHECKBOX, 'filter_show_limit' => 0];
                             $doneCategories[(int) $idCategory]['f' . (int) $featuresById[(int) $kFeature]] = true;
--- a/ps_facetedsearch.php
+++ b/ps_facetedsearch.php
@@ -403,7 +403,7 @@ class Ps_Facetedsearch extends Module implements WidgetInterface
         $shopList = Shop::getShops(false, null, true);
 
         foreach ($shopList as $idShop) {
-            $currencyList = Currency::getCurrencies(false, true);
+            $currencyList = Currency::getCurrencies(false, true, true);
 
             $minPrice = [];
             $maxPrice = [];
--- a/ps_facetedsearch.php
+++ b/ps_facetedsearch.php
@@ -403,7 +403,7 @@ class Ps_Facetedsearch extends Module implements WidgetInterface
         $shopList = Shop::getShops(false, null, true);
 
         foreach ($shopList as $idShop) {
-            $currencyList = Currency::getCurrencies(false, true, new Shop($idShop));
+            $currencyList = Currency::getCurrencies(false, true);
 
             $minPrice = [];
             $maxPrice = [];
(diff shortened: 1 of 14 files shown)


ps_facetedsearch.php
```diff
@@ -403,7 +403,7 @@ public function indexProductPrices($idProduct, $smart = true)
         $shopList = Shop::getShops(false, null, true);
 
         foreach ($shopList as $idShop) {
-            $currencyList = Currency::getCurrencies(false, true, new Shop($idShop));
+            $currencyList = Currency::getCurrencies(false, true);
```
jolelievre: Why is the shop condition removed here?
It will probably mess with the indexation no?

mattgoud: good catch @jolelievre. actually the 3rd arg is \`bool $groupBy\`, not a shop , the \`new Shop($idShop)\` was always truthy so it just forced GROUP BY on, it never scoped currencies by shop (that comes from the context shop in findAll). but dropping it entirely also dropped the GROUP BY, which dedupes currency rows in a shop-group context. so i kept it explicitly with \`true\`: \`getCurrencies(false, true, true)\` , same behavior as before, just a real bool for phpstan. fixed in 14ee4e1

jolelievre (approved):

mattgoud:
## 🧪 QA test plan

Most of the diff is dead-code removal / mechanical refactors, but a few changes touch runtime code. Here is what to focus on, by risk.

### 🔴 High — price indexation

`indexPrices()` was changed: `Currency::getCurrencies()` call signature + several `Product::getPriceStatic()` arguments switched from `null` to `0` / `''` (id_group, from_quantity…).

Re-run the price index, then check the **price-range filter (slider)** in FO:
- correct min/max on a category
- products with **specific prices / discounts**
- products with **group-specific prices**
- shop with **several currencies**
- **multistore** context (the loop iterates over shops)

### 🟠 Medium

- **BO forms** (attribute / attribute group / feature / feature value): `default_form_language` now goes through a typed helper. Edit a feature / attribute group in BO → language selector and multilingual display work.
- **FO assets** (`ProductSearch`): faceted block renders, price slider works with `PS_USE_JQUERY_UI_SLIDER` **on and off**, CSS/JS load.
- **Module config page (BO):** opens without error, drag & drop reordering of filters works (`Sortable`, `back.js`/`back.css`).

### 🟢 Low (mechanical refactors)

- **Save a filter template** with multistore shop association + attribute/feature filters → saved correctly.
- **Filter templates list:** date shown stays correct (`Tools::displayDate`).
- Navigation/pagination with active filters OK (`updateQueryString`).

### CI
The "PHP tests" workflow must stay green, including the new PHPStan jobs for 9.0.3 (PHP 8.1 & 8.4).

SiraDIOP (commented):
Hello @mattgoud, 

Thanks for your PR, i tested it and it's good for me :) 

Price indexation
Rebuilt the entire price index from the module config page → "Price indexing finished" displayed with no error.
FO price filter (slider)
Navigated to a category page (/3-clothes). The price slider renders correctly with a coherent min/max range (€22.00 – €44.00). Adjusted the slider to a narrower range (€23.49 – €29.36) → only the matching product was displayed, active filter shown with the correct range and working × to clear it. URL updated accordingly.

https://github.com/user-attachments/assets/52ae5b3e-f2a1-44dc-bd71-e28e22f8f022

Image

BO forms (attribute / attribute group / feature / feature value)
Edited attribute value "Gray" (Color group) and feature value "Polyester" (Composition). Language selector displays correctly in EN and FR on all multilingual fields (Name, URL, Meta title). Switching between languages updates the field content as expected. No error on form load or language switch.

Image

Image

Image
Image

FO assets — PS_USE_JQUERY_UI_SLIDER
Set PS_USE_JQUERY_UI_SLIDER=0 via console, then cleared the cache. On the category page (/3-clothes), a Smarty error is displayed:

offset on value of type null in /var/www/html/var/cache/dev/smarty/compile/hummingbird/.../2.module.ps_facetedsearchviewstemplatesfrontcatalogfacets.tpl.php on line 491

The error persists after cache clear and after setting PS_USE_JQUERY_UI_SLIDER back to 1. The faceted block and price slider are still visible but the error is shown on the page.
⚠️ To be investigated by the author — needs confirmation whether this is pre-existing or introduced by this PR.

Image
Image
Image
Image

Module config page (BO) — filter template
Opened the filter template edit form without error. Reordered filters via drag & drop (moved "Product brand filter" to the first position) and saved. The new order was correctly persisted after page reload.

https://github.com/user-attachments/assets/912de48e-936f-493b-be6b-fac3a29d76b2


FO navigation with active filters (updateQueryString)
Applied a Size: M filter on the category page — URL updated to ?q=Size-M, active filter tag displayed with working × to clear. Added a second filter (Color: White) — both filters remained active simultaneously (?q=Size-M/Color-White), active filter tags displayed correctly, product results updated accordingly.

Image
Image

Price indexation — multistore context
Rebuilt the entire price index from the module config page in "All stores" context (2 shops: PrestaShop on localhost:8001/ and test on localhost:8001/test/). Indexation completed successfully ("Price indexing finished"). Database check confirmed both shops have their prices correctly indexed (id_shop=1: 19 rows, id_shop=2: 19 rows in ps_layered_price_index).

Image

Price indexation — specific prices and group-specific prices
Added a group-specific price for the "Visitor" group on the Hummingbird printed t-shirt (−€10.00). Rebuilt the entire price index. In FO (unauthenticated, Visitor group), the product correctly shows the −€10.00 badge while the sweater retains its existing −20% specific price. The getPriceStatic() changes correctly handle group-specific pricing.

Image
Image

Filter template save and date display
Created a new filter template ("My template - 2026-06-25"). It appears correctly in the template list with the right creation date (06/25/2026 12:44:44). Tools::displayDate works as expected.

Image

SiraDIOP:
Summary
All scenarios from the test plan have been covered. The PHPStan / CI changes have no functional impact on the module's runtime behavior. Price indexation, FO price filter, BO multilingual forms, filter template management, drag & drop reordering, active filter navigation, multistore indexation, and group-specific pricing all work as expected.
One issue was identified during testing:
 A Smarty error is displayed on the FO category page regardless of the PS_USE_JQUERY_UI_SLIDER value:

offset on value of type null in .../2.module.ps_facetedsearchviewstemplatesfrontcatalogfacets.tpl.php on line 491

The error persists after cache clear. It needs to be confirmed whether this is pre-existing or introduced by this PR before merge.
**Cannot approve until the Smarty error is investigated.**

SiraDIOP:
@mattgoud While testing this PR, I noticed a Smarty error appearing on the front office category page when clicking the Price filter. The error occurs on develop but not on 9.1.x, which suggests it may be a regression introduced on develop and not necessarily caused by this PR itself.
I have opened a dedicated issue to track it: PrestaShop/PrestaShop#41846 
This does not block the validation of this PR since the error appears to be unrelated to the PHPStan changes introduced here.

PR #1255, opened 2026-06-24
Harden range filter labels against malformed values

Betafer:
## What

This PR makes range filter label generation resilient to malformed range values coming from the faceted search URL.

`SearchProvider::labelRangeFilters()` currently passes active range values directly to `formatNumber()` / `formatPrice()`. If a crafted URL provides a non-numeric range boundary, for example an HTML-like string, PrestaShop can throw a localization/decimal parsing exception while rendering category facets.

The change adds a small numeric boundary guard: valid numeric scalar values are preserved, invalid values fall back to the facet min/max, and a final `0` fallback is used only if both values are invalid.

## Why

This prevents category/listing pages from failing when bots or malicious requests submit malformed price/weight range boundaries.

## Tests

- `php -l src/Product/SearchProvider.php`
- `php -l tests/php/FacetedSearch/Product/SearchProviderTest.php`
- `PHP_CS_FIXER_IGNORE_ENV=1 vendor/bin/php-cs-fixer fix --no-interaction --dry-run --diff --config=.php_cs.dist src/Product/SearchProvider.php tests/php/FacetedSearch/Product/SearchProviderTest.php`
- Live validation on a PrestaShop 8 production/stage install: malformed `q=Prezzo-<a><a><a><a><a>-100` AJAX category requests now return HTTP 200 and valid rendered products/facets instead of throwing `cannot be interpreted as a number`.

Note: the full legacy PHPUnit 5 suite could not run in my local environment because it is incompatible with PHP 8.1 (`Cannot acquire reference to $GLOBALS`).

### DIFF
--- a/src/Product/SearchProvider.php
+++ b/src/Product/SearchProvider.php
@@ -442,8 +442,14 @@ class SearchProvider implements FacetsRendererInterface, ProductSearchProviderIn
 
             foreach ($facet->getFilters() as $filter) {
                 $filterValue = $filter->getValue();
-                $min = empty($filterValue[0]) ? $facet->getProperty('min') : $filterValue[0];
-                $max = empty($filterValue[1]) ? $facet->getProperty('max') : $filterValue[1];
+                $min = $this->getNumericRangeBoundary(
+                    isset($filterValue[0]) ? $filterValue[0] : null,
+                    $facet->getProperty('min')
+                );
+                $max = $this->getNumericRangeBoundary(
+                    isset($filterValue[1]) ? $filterValue[1] : null,
+                    $facet->getProperty('max')
+                );
                 if ($facet->getType() === 'weight') {
                     $unit = Configuration::get('PS_WEIGHT_UNIT');
                     $filter->setLabel(
@@ -466,6 +472,27 @@ class SearchProvider implements FacetsRendererInterface, ProductSearchProviderIn
                 }
             }
         }
+    }
+
+    /**
+     * Keep range labels resilient to malformed facet values coming from the URL.
+     *
+     * @param mixed $value
+     * @param mixed $fallback
+     *
+     * @return int|float|string
+     */
+    private function getNumericRangeBoundary($value, $fallback)
+    {
+        if (is_scalar($value) && is_numeric($value)) {
+            return $value;
+        }
+
+        if (is_scalar($fallback) && is_numeric($fallback)) {
+            return $fallback;
+        }
+
+        return 0;
     }
 
     /**
--- a/tests/php/FacetedSearch/Product/SearchProviderTest.php
+++ b/tests/php/FacetedSearch/Product/SearchProviderTest.php
@@ -411,4 +411,14 @@ class SearchProviderTest extends MockeryTestCase
             )
         );
     }
+
+    public function testNumericRangeBoundaryFallsBackWhenValueIsMalformed()
+    {
+        $method = new \ReflectionMethod(SearchProvider::class, 'getNumericRangeBoundary');
+        $method->setAccessible(true);
+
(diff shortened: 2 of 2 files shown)


ps-jarvis:
Hello @Betafer!

kpodemski:
> [!NOTE]

Thanks for digging into this — the exception you hit on PS 8 is real, and the write-up is clear. The problem is that on `dev` it has already been fixed upstream, and this guard doesn't behave the way it's meant to once that fix is in place.

### The root cause is already handled, in this PR's own base

cfadb10 *"Convert selected price and weight filters to floats"* (9 June, shipped in v5.0.0, and an ancestor of `fcc09ab`) sanitises both range filters at the boundary:

```php
$priceBlock['value']  = !empty($selectedFilters['price'])  ? array_map('floatval', $selectedFilters['price'])  : null;
$weightBlock['value'] = !empty($selectedFilters['weight']) ? array_map('floatval', $selectedFilters['weight']) : null;
```

`Converter` sets the filter value only from `$filterBlock['value']`, and `restorePriceAndWeightFilters()` touches the adapter rather than the block, so by the time `labelRangeFilters()` runs the value is always floats or `null` — never the string that made `formatPrice()` throw.

Replaying your payload against the base commit, unpatched:

| URL | base commit |
| --- | --- |
| `q=Price-€-<a><a><a><a><a>-100` | 200, label `€14.00 - €100.00` |
| `q=Price-€-abc-42` | 200, label `€14.00 - €42.00` |
| `q=Price-€-14-42` | 200, label `€14.00 - €42.00` |

No exception. Your PS 8 install would have been on a 3.x/4.x module, before cfadb10.

### With this patch the label gets slightly worse

Because the junk is already `floatval`'d to `0.0` before it reaches the guard, `is_numeric(0.0)` is true, so the new check accepts it and skips the fallback that `empty()` used to trigger:

| URL | base | this PR |
| --- | --- | --- |
| `q=Price-€-<a><a><a><a><a>-100` | `€14.00 - €100.00` | `€0.00 - €100.00` |
| `q=Price-€-abc-42` | `€14.00 - €42.00` | `€0.00 - €42.00` |
| `q=Price-€-14-42` | `€14.00 - €42.00` | `€14.00 - €42.00` |

So for exactly the malformed input this is meant to harden, the active-filter chip now advertises a €0.00 lower bound that is outside the facet's own range, where `dev` showed the facet minimum. Cosmetic and only reachable by hand-editing the URL, but it is a step backwards rather than forwards.

The guard can't really win here: sanitisation happens upstream, so by this point "the shopper picked 0" and "this was garbage" are the same value. To distinguish them the check would have to sit before the `floatval`, not after.

### On the test

`testNumericRangeBoundaryFallsBackWhenValueIsMalformed()` reaches through `ReflectionMethod` into a private helper, so it passes whether or not the real path is reachable — it would stay green even with `labelRangeFilters()` deleted. A test driving `labelRangeFilters()` with a malformed facet value would have surfaced the interaction above.

For what it's worth, the suite does run on PHP 7.4 (PHPUnit 5.7 predates PHP 8): `101 tests, 937 assertions` green on your branch.

### Where I land

I'm going to close this as superseded by cfadb10 — the crash it targets can't be reached on `dev` any more, and as it stands the patch trades that non-issue for a small regression in the rendered label.

Happy to reopen if you want to pursue the belt-and-braces version, but it would need to keep the current fallback behaviour: treat `0` as "not supplied" the way `empty()` did, or move the validation ahead of the `floatval` so it can still tell a real 0 from sanitised garbage. A test driving `labelRangeFilters()` rather than the private helper would need to come with it.

None of this is wasted, by the way — the exception you reported is real on 3.x/4.x, and if you are running the module on PS 8 it's worth knowing the fix is to move to 5.x rather than to patch the formatter.

PR #1258, opened 2026-06-27
Fix price indexation loop stopping early with non-sequential product ids

### LINKED ISSUE #1198: [BUG] Wrong loop break condition in indexPrices() prevents full price indexation
### Prerequisites

- [x] I have already [searched in existing issues](./issues?q=is%3Aissue+type%3ABug) and found no previous report of this bug.

When indexing product prices, the method indexPrices() uses an incorrect loop condition comparing $cursor (last indexed id_product) against $nbProducts (total number of products to index).

Because product IDs may not be sequential and are frequently greater than the total number of products, the loop exits after the first iteration, preventing full indexation.

### Expected behavior

**Actual result**

The loop stops because:
$cursor < $nbProducts

fails, since:
* $cursor is the last indexed product ID (example: 10391)
* $nbProducts is the number of products to index (example: 3525)

IDs are not sequential, therefore the check is invalid.

**Expected result**
The loop should continue until the number of indexed products reaches $nbProducts, not until $cursor is less than $nbProducts.

**Problematic code**
From indexPrices():
```
} while (
    $cursor < $nbProducts
    && (Tools::getMemoryLimit() == -1 || Tools::getMemoryLimit() > memory_get_peak_usage())
    && $time_elapsed < $maxExecutiontime
);
```

**Why this is a bug**
$cursor is the last product ID returned from: indexPricesUnbreakable().
It is **not** a counter.

Therefore comparing it to the number of products makes the loop stop incorrectly.

**Suggested fix**
Replace 
`$cursor < $nbProducts`

with:
`$indexedProducts < $nbProducts`

Since $indexedProducts already tracks real progress:
`$indexedProducts += $length;`

### Steps to reproduce

1. Install PrestaShop 8.x with ps_facetedsearch.
2. Have more product IDs than active products (normal in real stores: deleted products, migrations, imports).
3. Run price indexation with all products (full = true and default $length = 100).
4. The indexation stops after the first batch.

8.0.1

8.1.33

_No response_

boo-code:
| Questions         | Answers
| Description?      | `indexPrices()` used `$cursor` as the `do/while` continuation check: `$cursor < $nbProducts`. But `$cursor` is the **last indexed `id_product`** returned by `indexPricesUnbreakable()` (`return (int) $lastIdProduct;`), not a progress counter — while `$nbProducts` is the *number* of products to index. Product ids are not sequential and are routinely larger than the product count (gaps from deleted / imported / migrated products), so `$cursor < $nbProducts` flips to false as soon as an indexed id exceeds the count. The batching loop then stops after very few iterations and indexation falls back to a single 100-row batch per `indexPrices()` self-recursion (CLI/cron) or per AJAX round-trip (BO "rebuild index"). On large catalogues this both slows full indexation dramatically and deepens the `indexPrices()` recursion (one level per 100 products). The method already maintains `$indexedProducts += $length;`, which is the real progress counter, so the loop now compares against that instead.
| Type?             | bug fix
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | Fixes PrestaShop/ps_facetedsearch#1198.
| How to test?      | On a catalogue where the highest active `id_product` is greater than the number of indexable products (e.g. delete/migrate products so ids are sparse, or simply have more than ~100 products with non-contiguous ids), rebuild the faceted-search price index (BO > module configuration, or `Configuration::updateValue('PS_LAYERED_INDEXED', 0)` then trigger indexing). Before: the `layered_price_index` table is filled only in small steps (one 100-row batch per call), and on large/sparse catalogues full indexation is slow or doesn't complete within the recursion/limits. After: each call batches until the time/memory limit is reached and the table is fully populated.
| Sponsor company   |

#### Notes on verification

`indexPrices()`/`indexPricesUnbreakable()` are `private` methods on the main `Ps_facetedsearch` module class, which currently has no unit-test coverage (the suite targets the `src/` classes), and they are tightly coupled to the database. I verified the fix by tracing the code rather than with a mock-heavy reflection test: `indexPricesUnbreakable()` returns `$lastIdProduct` (an id, confirmed), and `$indexedProducts` increments by `$length` per batch, so `$indexedProducts < $nbProducts` is the correct continuation condition. It is monotonic (no infinite-loop risk) and the existing memory / `max_execution_time` / `$cursor == 0` guards are untouched. Happy to add a test if you have a preferred pattern for covering the module class.

### DIFF
--- a/ps_facetedsearch.php
+++ b/ps_facetedsearch.php
@@ -1583,7 +1583,11 @@ VALUES(' . $last_id . ', ' . (int) $idShop . ')');
             $time_elapsed = microtime(true) - $startTime;
             $indexedProducts += $length;
         } while (
-            $cursor < $nbProducts
+            // $cursor is the last indexed id_product (returned by indexPricesUnbreakable()), not a
+            // counter, so comparing it to the product count is wrong: with non-sequential ids (gaps
+            // from deleted/imported products) it stops batching as soon as an id exceeds $nbProducts.
+            // Track how many products were actually indexed instead.
+            $indexedProducts < $nbProducts
             && (Tools::getMemoryLimit() == -1 || Tools::getMemoryLimit() > memory_get_peak_usage())
             && $time_elapsed < $maxExecutiontime
         );


ps-jarvis:
Hello @boo-code!

kpodemski (approved):

Hlavtox:
@kpodemski @boo-code This is not precise. Indexed products vs total products doesn't necessarily mean that everything was indexed… in edge case.

if cursor < highestproduct id would be better.

kpodemski:
@Hlavtox i will take a look later, the simple test went well

boo-code:
@Hlavtox the count is imprecise, agreed, but completeness does not rest on it.

`$indexedProducts` is incremented by `$length` per batch (L1603), not by the rows the batch actually returned, so it overshoots by up to 99 on the final partial batch. That predates this change, only the comparison moved, and it is visible: the progress text sums `res.count` across calls, so a 250-product catalogue can finish reading `300/250`.

What it cannot do is end a run early. The do-while only bounds the work done in one request. Completion is cursor-driven in both paths:

- ajax: `_dev/back/index.js` re-clicks as long as the response carries a `cursor`, and the server answers `{'result': 'ok'}` only once a batch comes back empty. `$cursor == 0` at L1598 sets `$lastCursor = 0` as well, so `$cursor != $lastCursor` at L1618 is false and it falls through to the finished branch.
- non-ajax: the recursion at L1614 re-enters with the new cursor under the same condition.

So leaving the do-while early only ends that request, and the next one resumes from the cursor.

On `cursor < highest product id`: that is the right bound for the full path, which pages on `WHERE p.id_product > $cursor`. It does not carry over to the non-full path, whose query has no cursor clause at all. That one selects `WHERE psi.id_product IS NULL LIMIT 0,100` and re-reads the shrinking remainder on each pass, so the cursor does not track progress there and `$lastIdProduct` is just whatever the last row happened to be. A single condition for both would have to stay count-based on that side.

If the count is worth making exact, `indexPricesUnbreakable()` already walks the rows it indexed, so returning that number alongside the cursor and adding it instead of `$length` fixes both the loop bound and the progress text.

PR #1259, opened 2026-06-27
Always dispatch actionFacetedSearchFilters hook

### LINKED ISSUE #1239: actionFacetedSearchFilters hook is skipped when a "Categories" facet is active (early return in Search::addControllerSpecificFilters)
### Prerequisites

The actionFacetedSearchFilters hook — the documented extension point that lets a module add custom filters to the faceted-search adapter — is dispatched at the very end of   Search::addControllerSpecificFilters() (src/Product/Search.php). That method contains early return statements that skip the dispatch.

On a category page, as soon as the customer selects a "Categories" facet (so an id_category filter already exists), the method returns early and the hook is never fired:
 
private function addControllerSpecificFilters()
    {
        if ($this->query->getQueryType() == 'category') {
        
            if (!empty($this->getSearchAdapter()->getFilter('id_category'))) {
            
                return; // <-- early return: the Hook::exec below is never reached
        
            }
        
            // ...
    
        }
    
        // ...
        //Hook::exec('actionFacetedSearchFilters', ['search' => $this, 'query' => $this->query]);
    }

 Any module listening on actionFacetedSearchFilters is therefore silently bypassed whenever a "Categories" facet is active (same for the other early returns: new-products, prices-drop). The hook works fine for plain category pages, manufacturer, supplier, search, and feature/attribute facets, which makes the bug easy to miss. Confirmed by logging: initSearch() runs, but Hook::exec('actionFacetedSearchFilters', ...) is never reached.

### Steps to reproduce

1. Install a module that registers actionFacetedSearchFilters and adds a filter, e.g. $params['search']->getSearchAdapter()->addFilter('id_product', [1, 2, 3]);
2. Open a category page without a "Categories" facet → the hook fires, the custom filter is applied.
3. Select a "Categories" facet → id_category is now set → addControllerSpecificFilters() returns early → the hook is not dispatched → the custom filter is dropped.

### Expected behavior

actionFacetedSearchFilters should be dispatched on every faceted query, regardless of the early returns inside addControllerSpecificFilters(). A possible fix: move the Hook::exec('actionFacetedSearchFilters', ...) call out of addControllerSpecificFilters() and into initSearch() (right after $this->addControllerSpecificFilters();).

### Actual Result

When a "Categories" facet is active, modules hooked on actionFacetedSearchFilters are never called; their custom filters are silently ignored and do not appear in the final product SQL query.

9.1.1

4.0.4

8.1 (control-flow bug — reproducible on any supported PHP version)

_No response_

boo-code:
| Questions         | Answers
| Description?      | `actionFacetedSearchFilters` — the documented hook that lets a module add custom filters to the faceted-search adapter — was dispatched at the very end of `Search::addControllerSpecificFilters()` (`src/Product/Search.php`). That method contains several early `return` statements (category page when a "Categories" facet is already selected, `new-products` when a `date_add` filter is already set, `prices-drop` when a `reduction` filter is already set). Whenever one of those returns runs, the `Hook::exec()` line is never reached, so any module listening on `actionFacetedSearchFilters` is silently bypassed and its filters are dropped from the final product query. This is easy to miss because the hook works fine on plain category/manufacturer/supplier/search pages and on feature/attribute facets. Fix: move the `Hook::exec('actionFacetedSearchFilters', ...)` call out of `addControllerSpecificFilters()` and into `initSearch()`, right after `addControllerSpecificFilters()` — so it fires on every faceted query, while still running before the filters become the initial population.
| Type?             | bug fix
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | Fixes PrestaShop/ps_facetedsearch#1239.
| How to test?      | A unit test is included (`SearchTest::testInitSearchDispatchesFacetedSearchFiltersHookWithSelectedCategoryFacet`): on a category query, `initSearch(['category' => [[6]]])` sets an `id_category` filter (triggering the early return), and the test asserts `actionFacetedSearchFilters` is still dispatched — red before this change, green after. Manually: register a module on `actionFacetedSearchFilters` that calls `$params['search']->getSearchAdapter()->addFilter('id_product', [1, 2, 3]);`, open a category page, then select a "Categories" facet — before, the custom filter disappears from the results; after, it is applied in both cases.
| Sponsor company   |

### DIFF
--- a/src/Product/Search.php
+++ b/src/Product/Search.php
@@ -130,6 +130,18 @@ class Search
         // Adds filters that specific for this controller
         $this->addControllerSpecificFilters();
 
+        // Let modules add their own filters to the adapter. This is dispatched here, after the
+        // controller specific filters, rather than inside addControllerSpecificFilters() because
+        // that method has early returns (e.g. an already selected "Categories" facet) that would
+        // otherwise silently skip the hook and drop the module filters.
+        Hook::exec(
+            'actionFacetedSearchFilters',
+            [
+                'search' => $this,
+                'query' => $this->query,
+            ]
+        );
+
         // Add group by to remove duplicate values
         $this->getSearchAdapter()->addGroupBy('id_product');
 
@@ -451,14 +463,6 @@ class Search
                 empty($productPool) ? ['NULL'] : $productPool
             );
         }
-
-        Hook::exec(
-            'actionFacetedSearchFilters',
-            [
-                'search' => $this,
-                'query' => $this->query,
-            ]
-        );
     }
 
     /**
--- a/tests/php/FacetedSearch/Product/SearchTest.php
+++ b/tests/php/FacetedSearch/Product/SearchTest.php
@@ -1192,4 +1192,28 @@ class SearchTest extends MockeryTestCase
         $this->search->addFilter('weight', [10, 20]);
         $this->search->addFilter('id_feature', [[10, 20]]);
     }
+
+    /**
+     * The actionFacetedSearchFilters hook must be dispatched even when a controller specific
+     * filter causes an early return inside addControllerSpecificFilters(): on a category query,
+     * selecting a "Categories" facet sets an id_category filter and used to skip the hook.
+     *
+     * @see https://github.com/PrestaShop/ps_facetedsearch/issues/1239
+     */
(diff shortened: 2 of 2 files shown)


ps-jarvis:
Hello @boo-code!

kpodemski (approved):

PR #1260, opened 2026-06-27
Apply out-of-stock-last ordering on search result pages

### LINKED ISSUE #1236: Search relevance ordering bypasses out-of-stock-last logic
### Prerequisites

After updating `ps_facetedsearch`, search result pages no longer respect the "show out-of-stock products last" behavior.

The same stock ordering still works correctly on category, manufacturer/brand, and other product listing pages. The issue seems limited to search result pages.

### Steps to reproduce

1. Use PrestaShop 8.2 with PHP 8.1.
2. Install or update `ps_facetedsearch`.
3. Enable/configure the behavior that pushes out-of-stock products to the end.
4. Create or use a catalog where a search term returns both in-stock and out-of-stock products.
5. Run that search from the storefront.
6. Compare the result ordering with a category or brand listing.

### Expected behavior

When the "show out-of-stock products last" behavior is enabled, search results should list:

1. products with available stock first
2. products without stock at the end

The selected/default search ordering should still be preserved inside each group.

### Actual Result

On search result pages, out-of-stock products can appear before products that have stock.

8.2.1

4.0.4

8.1

Sercom AI

boo-code:
| Questions         | Answers
| Description?      | On search result pages the "show out of stock last" behaviour was never applied, while category, manufacturer and the other listings get it. Search results are ordered by relevance, which `Adapter\MySQL::computeOrderByField()` builds as a `FIELD(p.id_product, ...)` list from the product pool returned by the core search, and that branch **returned early**, before `computeShowLast()` was reached. Following the discussion below, search does not simply inherit the listing setting: it ranks by relevance rather than by browsing, so it gets **its own** setting, `PS_LAYERED_FILTER_SHOW_OUT_OF_STOCK_LAST_SEARCH`. A merchant can keep unavailable products last while browsing without changing what a search for a specific product returns, and the existing `PS_LAYERED_FILTER_SHOW_OUT_OF_STOCK_LAST` no longer affects search at all. Relevance is preserved within each availability group: the `FIELD()` ordering is applied after the in-stock split, so the most relevant available product leads and a relevant unavailable one drops below it rather than losing its place.
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | Fixes PrestaShop/ps_facetedsearch#1236.
| How to test?      | The new setting is "Show unavailable, out of stock last on search results" in the module configuration, and it defaults to **off**, so nothing changes until it is enabled. Two unit tests cover it in `MySQLTest`: `testGetQueryLeavesSearchUntouchedWhenOnlyTheListingSettingIsEnabled` asserts that turning the listing setting on leaves a search query as pure relevance (`FIELD(p.id_product,1,2,3) DESC`, and no `IFNULL(p.quantity, 0) <= 0`), and `testGetQueryAppliesOutOfStockLastOnSearchWhenTheSearchSettingIsEnabled` asserts that the new setting adds the out-of-stock-last clause while keeping the relevance ordering. Manually: enable only the new setting, run a storefront search returning both in- and out-of-stock products, and check unavailable ones move to the end; with only the old setting on, search is unchanged while category listings still order by availability.
| Sponsor company   |

The generated `ORDER BY` for a search query, by setting:

```
listing setting only : ORDER BY FIELD(p.id_product,1,2,3) DESC
search setting on    : ORDER BY IFNULL(p.quantity, 0) <= 0, IFNULL(p.quantity, 0) <= 0 AND FIELD(sa.out_of_stock, 1) DESC, FIELD(p.id_product,1,2,3) DESC
```

No upgrade script is needed: the key defaults to `0` on install and `Configuration::get()` returns a false-y value when it is absent, so shops upgrading without it behave exactly as they do today.

### DIFF
--- a/ps_facetedsearch.php
+++ b/ps_facetedsearch.php
@@ -208,6 +208,7 @@ class Ps_Facetedsearch extends Module implements WidgetInterface
             Configuration::updateValue('PS_ATTRIBUTE_ANCHOR_SEPARATOR', '-');
             Configuration::updateValue('PS_LAYERED_FILTER_PRICE_ROUNDING', 1);
             Configuration::updateValue('PS_LAYERED_FILTER_SHOW_OUT_OF_STOCK_LAST', 0);
+            Configuration::updateValue('PS_LAYERED_FILTER_SHOW_OUT_OF_STOCK_LAST_SEARCH', 0);
             Configuration::updateValue('PS_LAYERED_FILTER_BY_DEFAULT_CATEGORY', 0);
             Configuration::updateValue('PS_USE_JQUERY_UI_SLIDER', 1);
             Configuration::updateValue('PS_LAYERED_DEFAULT_CATEGORY_TEMPLATE', 0);
@@ -249,6 +250,7 @@ class Ps_Facetedsearch extends Module implements WidgetInterface
         Configuration::deleteByName('PS_LAYERED_FILTER_CATEGORY_DEPTH');
         Configuration::deleteByName('PS_LAYERED_FILTER_PRICE_ROUNDING');
         Configuration::deleteByName('PS_LAYERED_FILTER_SHOW_OUT_OF_STOCK_LAST');
+        Configuration::deleteByName('PS_LAYERED_FILTER_SHOW_OUT_OF_STOCK_LAST_SEARCH');
         Configuration::deleteByName('PS_LAYERED_FILTER_BY_DEFAULT_CATEGORY');
         Configuration::deleteByName('PS_LAYERED_FILTER_FEATURE_VALUES_USE_POSITION');
 
@@ -723,6 +725,7 @@ class Ps_Facetedsearch extends Module implements WidgetInterface
             Configuration::updateValue('PS_LAYERED_FILTER_CATEGORY_DEPTH', (int) Tools::getValue('ps_layered_filter_category_depth'));
             Configuration::updateValue('PS_LAYERED_FILTER_PRICE_ROUNDING', (int) Tools::getValue('ps_layered_filter_price_rounding'));
             Configuration::updateValue('PS_LAYERED_FILTER_SHOW_OUT_OF_STOCK_LAST', (int) Tools::getValue('ps_layered_filter_show_out_of_stock_last'));
+            Configuration::updateValue('PS_LAYERED_FILTER_SHOW_OUT_OF_STOCK_LAST_SEARCH', (int) Tools::getValue('ps_layered_filter_show_out_of_stock_last_search'));
             Configuration::updateValue('PS_LAYERED_FILTER_BY_DEFAULT_CATEGORY', (int) Tools::getValue('ps_layered_filter_by_default_category'));
             Configuration::updateValue('PS_USE_JQUERY_UI_SLIDER', (int) Tools::getValue('ps_use_jquery_ui_slider'));
             Configuration::updateValue('PS_LAYERED_DEFAULT_CATEGORY_TEMPLATE', (int) Tools::getValue('ps_layered_default_category_template'));
@@ -822,6 +825,7 @@ class Ps_Facetedsearch extends Module implements WidgetInterface
             'limit_warning' => $this->displayLimitPostWarning(21 + count($attributeGroups) * 3 + count($features) * 3),
             'price_use_rounding' => (bool) Configuration::get('PS_LAYERED_FILTER_PRICE_ROUNDING'),
             'show_out_of_stock_last' => (bool) Configuration::get('PS_LAYERED_FILTER_SHOW_OUT_OF_STOCK_LAST'),
+            'show_out_of_stock_last_search' => (bool) Configuration::get('PS_LAYERED_FILTER_SHOW_OUT_OF_STOCK_LAST_SEARCH'),
             'filter_by_default_category' => (bool) Configuration::get('PS_LAYERED_FILTER_BY_DEFAULT_CATEGORY'),
             'use_jquery_ui_slider' => (bool) Configuration::get('PS_USE_JQUERY_UI_SLIDER'),
             'default_category_template' => Configuration::get('PS_LAYERED_DEFAULT_CATEGORY_TEMPLATE'),
--- a/src/Adapter/MySQL.php
+++ b/src/Adapter/MySQL.php
@@ -36,6 +36,21 @@ class MySQL extends AbstractAdapter
     const TYPE = 'MySQL';
 
     /**
+     * Pushes out-of-stock products last on category, manufacturer and the other listings.
+     *
+     * @var string
+     */
+    const CONFIG_SHOW_OUT_OF_STOCK_LAST = 'PS_LAYERED_FILTER_SHOW_OUT_OF_STOCK_LAST';
+
+    /**
+     * Same behaviour on search result pages. Kept separate because search ranks by relevance,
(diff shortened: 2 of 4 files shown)


ps-jarvis:
Hello @boo-code!

kpodemski (dismissed):

kpodemski:
Hello @boo-code 

CI is red.

```
PHPUnit 5.7.27 by Sebastian Bergmann and contributors.

.E.....................................EE......................  63 / 101 ( 62%)
......................................                          101 / 101 (100%)

Time: 194 ms, Memory: 16.00MB

There were 3 errors:

1) PrestaShop\Module\FacetedSearch\Tests\Adapter\MySQLTest::testGetQueryAppliesOutOfStockLastOnSearchRelevanceOrdering
Error: Call to undefined method PrestaShop\Module\FacetedSearch\Tests\Adapter\MySQLTest::assertStringContainsString()

/home/runner/work/ps_facetedsearch/ps_facetedsearch/tests/php/FacetedSearch/Adapter/MySQLTest.php:111
phpvfscomposer:///home/runner/work/ps_facetedsearch/ps_facetedsearch/vendor/phpunit/phpunit/phpunit:51

2) PrestaShop\Module\FacetedSearch\Tests\Adapter\MySQLTest::testGetQueryWithComputeShowLastEnabled
Mockery\Exception\RuntimeException: Could not load mock Product, class already exists

/home/runner/work/ps_facetedsearch/ps_facetedsearch/vendor/mockery/mockery/library/Mockery/Container.php:218
/home/runner/work/ps_facetedsearch/ps_facetedsearch/vendor/mockery/mockery/library/Mockery.php:171
/home/runner/work/ps_facetedsearch/ps_facetedsearch/tests/php/FacetedSearch/Adapter/MySQLTest.php:455
phpvfscomposer:///home/runner/work/ps_facetedsearch/ps_facetedsearch/vendor/phpunit/phpunit/phpunit:51

3) PrestaShop\Module\FacetedSearch\Tests\Adapter\MySQLTest::testGetQueryWithComputeShowLastEnabledAndDenyOrderOutOfStockProducts
Mockery\Exception\RuntimeException: Could not load mock Product, class already exists

/home/runner/work/ps_facetedsearch/ps_facetedsearch/vendor/mockery/mockery/library/Mockery/Container.php:218
/home/runner/work/ps_facetedsearch/ps_facetedsearch/vendor/mockery/mockery/library/Mockery.php:171
/home/runner/work/ps_facetedsearch/ps_facetedsearch/tests/php/FacetedSearch/Adapter/MySQLTest.php:479
phpvfscomposer:///home/runner/work/ps_facetedsearch/ps_facetedsearch/vendor/phpunit/phpunit/phpunit:51

ERRORS!
Tests: 101, Assertions: 917, Errors: 3.
Script @php -d date.timezone=UTC ./vendor/bin/phpunit -c tests/php/phpunit.xml handling the test event returned with error code 2
```

boo-code:
Thanks. Fixed in 1d00b91. The new test used `assertStringContainsString()` (not available in PHPUnit 5.7) and declared a global `Product` stub in MockProxy, which stopped the existing computeShowLast tests from creating their `Mockery::namedMock(Product::class)` (class already exists). It now mocks Product through `namedMock()` like the neighbouring tests and asserts with `assertContains()`. PHPUnit is green again.

Hlavtox (changes requested):
Hello @boo-code, I have to disagree on this one, because I think it's the expected behavior.

Let me explain. When you search for anything, you would still expect the most relevant thing to be in the front, even if it's not in stock.

boo-code:
Thanks @Hlavtox, that is a fair point and I am happy to defer to your call on the module's direction. Two clarifications on what the PR actually does, in case they shift the picture:

- It only reorders when the merchant has explicitly enabled `PS_LAYERED_FILTER_SHOW_OUT_OF_STOCK_LAST`. With that setting off, search is untouched and stays pure relevance.
- Even with it on, relevance is preserved within each stock group: the `FIELD(p.id_product, ...)` relevance order is kept and applied after the in-stock / out-of-stock split, so the most relevant in-stock result still leads and a highly relevant out-of-stock item only drops below the in-stock ones rather than losing its relative order.

The intent was consistency: category, manufacturer and other listings already route through `computeShowLast` and honour that setting, and search was the one surface that returned early and skipped it. But if you would still prefer search to stay relevance-first regardless of the setting, I am glad to close this.

kpodemski:
cc @Hlavtox

kpodemski:
ping @Hlavtox :)

Hlavtox:
@kpodemski I still disagree, in case of search, this setting should be kept ignored imho.

kpodemski:
@Hlavtox 

configurable, maybe?

boo-code:
@Hlavtox , what if all products in a first search results page are out of stock? Visitor will leave the site. It happens for shops with large inventory and long history.

kpodemski:
@Hlavtox I agree with @boo-code reasoning, and I think we should at least have a way to enforce this behavior

Hlavtox:
@boo-code @kpodemski Guys, if you make this configurable, then I'm fine with it. 👍

Personally though, I definitely wouldn't enable it on any of my stores.

I see category listings and search results as two different use cases. In a category, the customer is usually browsing and choosing between multiple products, so pushing available products to the top can make a lot of sense.

With search, however, the customer is often looking for something specific. In that case I would expect relevance to remain the strongest sorting signal. If the best match happens to be out of stock, I would still rather show it first than put a less relevant product above it just because it is available.

So I'm absolutely OK with adding an option for merchants who prefer availability-first search, I just wouldn't make the existing "show out-of-stock products last" setting automatically affect search results as well.

kpodemski:
Let's do it then @boo-code :)

ps-jarvis:

    - [ ] `Show unavailable, out of stock last on search results`
    - [ ] `Search results are ordered by relevance. Enable this to push unavailable products to the end there as well.`

boo-code:
Done, and made configurable as agreed.

Search no longer follows the listing setting at all. It gets its own, `PS_LAYERED_FILTER_SHOW_OUT_OF_STOCK_LAST_SEARCH`, shown in the module configuration as "Show unavailable, out of stock last on search results" and defaulting to **off** — so this changes nothing on any shop until a merchant turns it on, and @Hlavtox's stores keep relevance-first search whatever the listing setting is doing.

Two tests, one for each half of that:

- `testGetQueryLeavesSearchUntouchedWhenOnlyTheListingSettingIsEnabled` — listing setting on, search setting off: the query is pure relevance, `FIELD(p.id_product,1,2,3) DESC`, with no `IFNULL(p.quantity, 0) <= 0`. This is the one that pins the boundary; gating search on the old key instead turns it red.
- `testGetQueryAppliesOutOfStockLastOnSearchWhenTheSearchSettingIsEnabled` — search setting on: the out-of-stock-last clause appears and the relevance ordering is kept, so within each availability group the most relevant product still leads.

Relevance stays the ordering inside each group rather than being replaced, so a highly relevant unavailable product drops below the available ones instead of losing its position among them.

Implementation is one optional argument on `computeShowLast()` for the setting to read, so both paths share the same code. No upgrade script: the key defaults to `0` on install and `Configuration::get()` is false-y when absent, so an existing shop that has not got the key behaves exactly as it does today.

Rebased onto current `dev` and squashed to a single commit. Locally 119 tests / 1089 assertions pass, and the description has been rewritten to match this design.

kpodemski (approved):

PR #1261, opened 2026-06-27
Index product prices for every active shop country

### LINKED ISSUE #1206: [BUG] No product at all for foreign countries when order by price
### Prerequisites

- [x] I have already [searched in existing issues](./issues?q=is%3Aissue+type%3ABug) and found no previous report of this bug.

If your client is from a foreign country, and choose to sort by price, the search return 0 products.

### Expected behavior

It should display the same product list as any other sorting.

### Steps to reproduce

Create a customer from an other country (my country is France, issues is for Luxemburg, Belgium etc) and try to sort by price asc.

8.2.3

8.2

_No response_

boo-code:
| Questions         | Answers
| Description?      | When a customer's country has no specific tax rule for a product, sorting search/listing results **by price** returned **0 products** for that customer, while every other sort worked. Root cause: `indexProductPrices()` computes prices for all countries but only **inserts** `layered_price_index` rows for the countries returned by the tax-rule query (`$taxRatesByCountry`) when `PS_LAYERED_FILTER_PRICE_USETAX` is on. The price sort joins that table with `INNER JOIN ... AND psi.id_country = <customer country>`, so a customer from a country that was never indexed matches no rows and gets an empty list. The fix indexes **every active shop country**: countries without a specific tax rule are indexed untaxed (0% rate) — exactly what the module already does for products that have no tax rule at all. Taxed countries keep their computed prices unchanged.
| Type?             | bug fix
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | Fixes PrestaShop/ps_facetedsearch#1206.
| How to test?      | On a shop with at least two active countries where products have a tax rule for only one of them (e.g. FR), log in / set the context to the other country (e.g. US), open a listing and sort by price — before, the list is empty; after re-indexing prices, it shows the products. Verified on the default shop (active countries FR=8, US=21; products taxed for FR only): re-indexing product 1 keeps FR identical (`price_min=22.944, price_max=28.680`) and adds the US row untaxed (`price_min=19.120, price_max=23.900`), and the price-sort `INNER JOIN` on `id_country=21` now returns the product instead of nothing.
| Sponsor company   |

Verification (default shop, `PS_LAYERED_FILTER_PRICE_USETAX` on):

```
before:  layered_price_index for product 1 -> only id_country=8 (FR): 22.944 / 28.680
         price sort for US (id_country=21)  -> 0 products

after:   id_country=8  (FR): 22.944 / 28.680   (unchanged)
         id_country=21 (US): 19.120 / 23.900   (untaxed, newly indexed)
         price sort for US (id_country=21)  -> returns the product
```

`indexProductPrices()` is a private method on the module class (no unit-test harness covers it, like the rest of the indexer), so this was verified by re-indexing against a real database and diffing the rows, confirming the taxed-country prices are byte-identical. The index now covers active **shop** countries, with the same row shape the no-tax-rule path already produces.

### DIFF
--- a/ps_facetedsearch.php
+++ b/ps_facetedsearch.php
@@ -424,18 +424,27 @@ class Ps_Facetedsearch extends Module implements WidgetInterface
                 'GROUP BY id_product, tr.id_country'
             );
 
-            if (empty($taxRatesByCountry) || !Configuration::get('PS_LAYERED_FILTER_PRICE_USETAX')) {
-                $shopCountries = Country::getCountriesByIdShop($idShop, $this->getContext()->language->id);
-                $taxCountries = array_filter($shopCountries, function ($country) {
-                    return $country['active'];
-                });
-                $taxRatesByCountry = array_map(function ($country) {
-                    return [
+            // When tax is not applied to indexed prices, drop the per-country tax rates entirely.
+            if (!Configuration::get('PS_LAYERED_FILTER_PRICE_USETAX')) {
+                $taxRatesByCountry = [];
+            }
+
+            // Always index every active country of the shop. Countries that have no specific tax
+            // rule for this product (or when the "use tax" option is off) are indexed with a 0% rate.
+            // Without this, the price sort - an INNER JOIN on layered_price_index.id_country - returns
+            // no products at all for customers whose country was not indexed. This makes the
+            // with-tax-rules path consistent with the path already used for products without any tax
+            // rule. See https://github.com/PrestaShop/ps_facetedsearch/issues/1206
+            $indexedCountryIds = array_column($taxRatesByCountry, 'id_country');
+            $shopCountries = Country::getCountriesByIdShop($idShop, $this->getContext()->language->id);
+            foreach ($shopCountries as $country) {
+                if (!empty($country['active']) && !in_array($country['id_country'], $indexedCountryIds)) {
+                    $taxRatesByCountry[] = [
                         'rate' => 0,
                         'id_country' => $country['id_country'],
                         'iso_code' => $country['iso_code'],
                     ];
-                }, $taxCountries);
+                }
             }
 
             $productMinPrices = $this->getDatabase()->executeS(


ps-jarvis:
Hello @boo-code!

kpodemski (approved):

kpodemski (approved):
> [!NOTE]

Reproduced and confirmed. No rebase needed, merges clean.

My default shop happened to be exactly your scenario already — active countries FR (8) and US (21), products taxed for FR only, and `layered_price_index` covering only country 8. So the bug was sitting there untouched.

### Before

With the context country set to US, a full re-index on `dev` still produced only country-8 rows, and the listing behaved exactly as #1206 describes — every other sort fine, price sort empty:

```
position sort -> 11 products
name sort     -> 11 products
price asc     ->  0 products
price desc    ->  0 products
```

us-before-empty

Worth noting the price *facet* disappears too, not just the sort — same INNER JOIN, same root cause.

### After

us-after-11-products

Both directions come back and are correctly ordered, and the slider reads 11–35 (untaxed US) against 14–42 (taxed FR), which is coherent.

### The claim I most wanted to check

That taxed countries are untouched. I snapshotted the whole table, re-indexed all 19 products on your branch, and diffed:

```
country 8 rows: IDENTICAL to the pre-change snapshot
country 21 rows: 19 added
product 1 -> FR 22.944 / 28.680 (unchanged), US 19.120 / 23.900 (new)
```

Byte-identical, and your quoted figures for product 1 match to the last decimal. Switching the context back to FR gives the same 11 products at €14.28 with the same 14–42 slider as before the change, so this is purely additive.

The 0% rate is also the right value rather than just a safe one: a tax rules group with no rule for a country means no tax there, and the US listing renders €11.90 where FR renders €14.28 — the indexed row agrees with what the shop actually charges.

Tests pass (100 / 929).

### Two notes, neither blocking

- `Country::getCountriesByIdShop()` has no cache and is now called unconditionally, so a full re-index runs one extra query per product per shop where it previously only ran for products with no tax rule. Harmless on a small catalog, less so on the 14k-product shops people bring to this module — it is shop-scoped and loop-invariant, so hoisting or memoising it would be an easy win.
- This pulls in the opposite direction from #1226, which wants to *shrink* the index by dropping countries. Correctness first, so this one should land regardless, but the two want sequencing and #1226 will need rework on top of it.

LGTM.

boo-code:
Confirmed at source level, and it is a regression this PR introduced rather than a pre-existing cost.

`Country::getCountriesByIdShop()` was previously reached only inside `if (empty($taxRatesByCountry) || !Configuration::get('PS_LAYERED_FILTER_PRICE_USETAX'))`. It is now called on every `indexProductPrices()`, so once per product per shop on a full re-index. Core's implementation carries no cache at all, a plain `executeS` with three joins on each call, so a 14k-product catalogue pays 14k extra queries. The result is shop-scoped and loop-invariant, so hoisting it out of the per-product path or memoising per `(idShop, idLang)` removes all of it.

I can fix it, but I have no free PR slots for the correction, so I could only open it as a draft. Worth being explicit that a draft does not help with the quota either: this repo's workflows are `on: [push, pull_request]` with no draft guard, so a draft runs the full matrix exactly like a ready PR, 17 check-runs on my current one. Tell me which you prefer, or whether it is better folded into an open PR that already touches the indexer.

Hlavtox:
cc @kpodemski check @boo-code comment

kpodemski:
@boo-code you can open a PR that will be a draft, as you see we're trying to move forward with your contributions ;)

boo-code:
Opened as a draft: #1310.

It covers both lookups in that loop, not only the one this PR made unconditional. `Country::getCountries()` a few lines below has the same shape and was already being called once per product per shop, and neither is cached in core.

Measured on a default 9.2 install (19 products, 2 shops), instrumenting both call sites and re-indexing every product:

```
before:  getCountriesByIdShop_calls=38  getCountries_calls=38
after:   getCountriesByIdShop_calls=2   getCountries_calls=1
```

2 and 1 are the expected cardinalities, one per shop-and-language and one per language. The `layered_price_index` rows are byte identical between the two runs, so it only removes repeated queries. For a catalogue of N products over S shops it goes from 2 x N x S of these queries to S + 1.

One sequencing note: it touches `indexProductPrices()`, which is also where #1226 works, so whichever lands second will need a rebase there.

boo-code:
Good to hear, and thanks for picking these up.

There is a plan behind the volume, so it is probably worth stating it once. I have been working the old bug backlog: issues reported years ago that still reproduce on the current version. Each one gets checked against 9.x before anything is written, and plenty were dropped at that step because they no longer reproduce.

I went newest to oldest. The June batch was 2019-2022, the end of July reached 2018-09, and the last batch on 10 August got to the oldest cohort in the tracker, 22 August 2018 — the `[BOOM-*]` issues migrated from the old system. #9787, #9817 and #9871 are from that day, fixed by #42327, #42329 and #42324. That is where it stopped, because I ran into the 10 open pull request limit for contributors without write access.

The remaining tail is shorter than it looks: 20 open issues are older than 23 August 2018, and 47 in total were reported that year. I will work through what is left as slots free up, and I am in no hurry — reviewing them properly matters more than opening more.

Between the backlog shrinking and everything you have merged this week, PrestaShop could end up with a remarkably short list of long-standing bugs for a project this size. That is the goal worth the effort.

PR #1276, opened 2026-07-28
List a product on the page of every supplier it is associated with

### LINKED ISSUE #41531: Supplier page only shows products where the supplier is the default supplier, ignoring ps_product_supplier associations
### Prerequisites

- [x] I have already [searched in existing issues](https://github.com/PrestaShop/PrestaShop/issues?q=is%3Aissue%20type%3ABug) and found no previous report of this bug.

PrestaShop 9.1.0 Module version: v4.0.3
When a product is associated with multiple suppliers via the back office (which writes rows to ps_product_supplier), the front-office supplier page only lists the product under the supplier marked as default (ps_product.id_supplier). The product is missing from the FO page of every other associated supplier.
This contradicts the back-office spec, which states that any supplier checked on the product's "Suppliers" tab should display the product on its front-office page.

### Steps to reproduce

1. Fresh PrestaShop 9.1 install with ps_facetedsearch enabled.
2. Create two suppliers, A and B. Make sure both are active and associated with the shop.
3. Create a product. In the product's Suppliers tab, check both A and B. Set A as default.
4. Save the product, rebuild the faceted search index from BO → Modules → Faceted Search → Configure.
5. Visit the front-office supplier page for A → product is listed.
6. Visit the front-office supplier page for B → product is missing.
7. Disable ps_facetedsearch and reload B's page → product appears.

### Expected behavior

A product associated with suppliers A and B should appear on both supplier A's and supplier B's front-office pages.
This matches what PrestaShop core does since Supplier::getProducts() was rewritten in PS 9 to query ps_product_supplier rather than ps_product.id_supplier:

// classes/Supplier.php, getProducts()
JOIN `'. _DB_PREFIX_ .'product_supplier` ps ON (ps.id_product = p.id_product)
...
WHERE ps.`id_supplier` = ' . (int) $idSupplier

So when ps_facetedsearch is disabled, both supplier pages list the product correctly.

### Actual Result

With ps_facetedsearch enabled, the module's search provider intercepts the supplier page (via the productSearchProvider hook) and runs its own filter:
// modules/ps_facetedsearch/src/Product/Search.php, line 388
// Supplier controller
if ($this->query->getQueryType() == 'supplier') {
    $this->getSearchAdapter()->addFilter('id_supplier', [$this->query->getIdSupplier()]);
}
id_supplier here resolves to the column on ps_product, which only stores the default supplier. Any product associated with a supplier via ps_product_supplier — but where that supplier is not the default — is filtered out.

9.1.0

### How have you installed PrestaShop

downloading

8.4.11

_No response_

boo-code:
Fixes PrestaShop/PrestaShop#41531

### Problem

A product associated with several suppliers through the product's Suppliers tab only appears on the front-office page of the supplier marked as default. The association written to `product_supplier` is ignored.

`Search::initSearch()` filters the supplier page with `addFilter('id_supplier', ...)`, but `id_supplier` has no entry in `MySQL::getFieldMapping()`. Fields absent from that map are resolved against the base product table, so the condition becomes `p.id_supplier`, which holds only the default supplier:

```sql
SELECT p.id_product FROM ps_product p WHERE p.id_supplier='7' ORDER BY p.id_product DESC
```

### Fix

Map `id_supplier` to `product_supplier`, the table the association actually lives in:

```sql
SELECT p.id_product FROM ps_product p
  INNER JOIN ps_product_supplier psup ON (p.id_product = psup.id_product)
  WHERE psup.id_supplier='7' ORDER BY p.id_product DESC
```

Both queries are the real output of `MySQLTest`, before and after.

### Why the join is safe

`product_supplier` holds a row per product, combination and supplier, so a join can repeat a product. Both consumers already collapse that:

- the listing groups by product - `Search.php` calls `addGroupBy('id_product')`, and so does `Filters\Products`,
- the counts use `COUNT(DISTINCT p.id_product)` in `valueCount()`, which `count()` delegates to.

The mapping also cannot affect other pages: `computeJoinConditions()` builds joins only from the select fields and the filter keys in use, so the join appears only when `id_supplier` is actually filtered, which is the supplier controller. `psup` is used as the alias because `ps` is already taken by `product_shop`.

### Tests

`MySQLTest::testGetQueryWithSupplierFilterUsesTheAssociationTable` asserts the generated SQL. Reverting only `src/Adapter/MySQL.php` fails it with the `p.id_supplier` query above. The full suite passes: 101 tests, 933 assertions.

php-cs-fixer with `.php_cs.dist` reports nothing to fix. I could not run PHPStan locally - it is not in the module's `vendor/`, it comes from the shared `php-ci/phpstan` action - so that one is left to CI. The change is a literal array entry, so there is little for it to chew on.

### DIFF
--- a/src/Adapter/MySQL.php
+++ b/src/Adapter/MySQL.php
@@ -220,6 +220,15 @@ class MySQL extends AbstractAdapter
                 'joinCondition' => '(p.id_product = cp.id_product)',
                 'joinType' => self::INNER_JOIN,
             ],
+            // The supplier page lists every product associated with the supplier, and that
+            // association lives in product_supplier. Without this mapping the filter falls back to
+            // p.id_supplier, which only holds the default supplier of the product.
+            'id_supplier' => [
+                'tableName' => 'product_supplier',
+                'tableAlias' => 'psup',
+                'joinCondition' => '(p.id_product = psup.id_product)',
+                'joinType' => self::INNER_JOIN,
+            ],
             'manufacturer_name' => [
                 'tableName' => 'manufacturer',
                 'tableAlias' => 'm',
--- a/tests/php/FacetedSearch/Adapter/MySQLTest.php
+++ b/tests/php/FacetedSearch/Adapter/MySQLTest.php
@@ -330,6 +330,22 @@ class MySQLTest extends MockeryTestCase
     }
 
     /**
+     * A product is listed on the page of every supplier it is associated with, and that association
+     * lives in product_supplier. Filtering on the product's own id_supplier column would only match
+     * the default supplier.
+     */
+    public function testGetQueryWithSupplierFilterUsesTheAssociationTable()
+    {
+        $this->adapter->setSelectFields(['id_product']);
+        $this->adapter->addFilter('id_supplier', [7], '=');
+
+        $this->assertEquals(
+            'SELECT p.id_product FROM ps_product p INNER JOIN ps_product_supplier psup ON (p.id_product = psup.id_product) WHERE psup.id_supplier=\'7\' ORDER BY p.id_product DESC',
+            $this->adapter->getQuery()
+        );
+    }
+
+    /**
      * @dataProvider getManyOperationsFilters
      */
     public function testGetQueryWithManyOperationsFilters($fields, $operationsFilter, $expected)


Touxten:
Close and open for news tests

kpodemski (approved):

kpodemski (approved):
Tested on PS 9.2, works as described. 101 tests / 933 assertions pass, and reverting only `src/Adapter/MySQL.php` makes the new test fail with the `p.id_supplier` query.

Demo data already covers the case: products 3/4/5 are linked to supplier 2 in `product_supplier` but default to supplier 1.

**Before** — supplier page says 14:
before-14-products

**After** — 17, posters now listed:
after-17-products

17 / 8 is also what core's supplier list showed all along, so this removes a real inconsistency:
core-supplier-list-17-8

Fashion supplier goes 5 → 8 (ids 1,2,3,4,5,12,13,14); nothing is lost on either page.

One note, not a blocker: the `INNER JOIN` means a product with `product.id_supplier` set but no matching `product_supplier` row would drop off that supplier's page. Normal BO flows keep those in sync and my DB had 0 such rows, but it's a behaviour change for shops with imported data.

boo-code:
On the `INNER JOIN` note, agreed it is a behaviour change and not only a fix. A product with `product.id_supplier` set and no matching `product_supplier` row is now absent from that supplier's page, where before it was listed. Normal BO flows keep the two in sync, so a clean shop shows nothing, but imports and direct SQL writes do not, and there the product disappears from the listing with no error to point at.

If it is worth guarding, the mapping can keep the association join and fall back to `p.id_supplier` for products with no association row, at the cost of a slightly heavier query.

Same constraint as on #1261: I can write it, but I have no free PR slots for the correction and could only open it as a draft, and a draft saves nothing here since the workflows are `on: [push, pull_request]` with no draft guard and run the full matrix regardless.

PR #1278, opened 2026-08-03
Pin PrestaShop 9.1.4 in CI instead of 9.1.x branch

kpodemski:
| Questions         | Answers
| Description?      | Points the PHPStan CI matrix at the PrestaShop `9.1.4` tag instead of the `9.1.x` branch, which is about to be closed. Also renames `tests/php/phpstan/phpstan-9.1.x.neon` to `phpstan-9.1.4.neon`, because the workflow derives the PHPStan config filename from the matrix value (`phpstan-config: tests/php/phpstan/phpstan-${{ matrix.presta_version }}.neon`) — without the rename the job would fail on a missing config. This matches the existing `phpstan-9.0.3.neon`, which is already pinned to a tag.
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | n/a
| How to test?      | Check this PR's "PHP tests" workflow run: the two `PHPStan (PHP 8.1 - 8.5)` jobs that previously ran against `9.1.x` must now show `9.1.4` in their matrix names and be green. No job should reference `9.1.x` any more. The `9.0.3`, `develop` and `8.2.x` jobs are untouched.
| Sponsor company   | n/a

Only CI configuration is affected — no module code changes.

The `9.1.x` string still appears once, in a comment that describes which PHP versions the branch family supports; it has been reworded to `# PrestaShop 9.1.x (pinned to the 9.1.4 tag) and develop support PHP 8.1 up to 8.5` so it stays accurate.


mattgoud:
@kpodemski heads up: this one is superseded twice over now. #1285 already replaced the `9.1.x` branch with the `9.1.4` tag, and #1301 moves it on to `9.1.5` (the last release of that line) while adding the `9.2.x` jobs this module was missing.

Fine to close whenever you want.

kpodemski:
@mattgoud if you such situation feel free to close the PR 👍🏻 thanks

PR #1283, opened 2026-08-04
fix: re-check price for product whose indexed range exceeds the filter

### LINKED ISSUE #1282: Price filter: products escape the price re-check when their indexed price_min equals the upper bound
### Prerequisites

`Products::filterPrice()` decides whether a product's real price needs to be recomputed with this test:

```php
if (($product['price_min'] < (int) $priceFilter['min'] && $product['price_max'] > (int) $priceFilter['min'])
|| ($product['price_max'] > (int) $priceFilter['max'] && $product['price_min'] < (int) $priceFilter['max']) 
) { 
``` 

The SQL that selected the product used **inclusive** bounds (`Search::addPriceFilter()`):

```php
$this->getSearchAdapter()->addFilter('price_min', [$maxPrice], '<='); 
$this->getSearchAdapter()->addFilter('price_max', [$minPrice], '>='); 
``` 

A product whose indexed `price_min` is exactly equal to the requested upper bound therefore passes the SQL filter (`price_min <= max`) but fails the second clause of the PHP
test (`price_min < max`). It is never re-checked, and it stays in the listing at whatever its real price is. The mirror case exists on the lower bound, when `price_max` equals `min`.

Concrete case on our shop:

| | value | 
| --- | --- | 
| indexed range | `30.00` -> `75.00` | 
| real price for a guest | `75.00` |
| requested filter | `3` -> `30` | 
| clause 1 | `30 < 3` -> false | 
| clause 2 | `75 > 30` **and** `30 < 30` -> false |

The 75 CHF product is listed under a "up to 30" filter.

The clause `price_min < max` is not needed at all. What the test wants to know is whether the indexed envelope sticks out of the requested range : if the envelope is fully contained, the real price is necessarily in range and no recompute is needed. The current test is a strict subset of that, and the products it misses are exactly the ones sitting on a bound.

### Steps to reproduce

1. PrestaShop 8.x, `ps_facetedsearch` enabled with a price filter on a category.
2. Take a product priced **75.00 tax incl.** and give the *Customer* group a **60 % discount** (Customers > Groups). The indexed envelope becomes `30.00 -> 75.00`.
3. Rebuild the price index
4. As a **guest**, filter that category on the price range **3 -> 30**.

The 75.00 product is listed. Any other upper bound (29, 31…) excludes it correctly, only the bound that lands exactly on the indexed `price_min` triggers it.

### Expected behavior

A product is listed under a price filter only if its real price for the current visitor is inside the requested range, regardless of where its indexed envelope happens to fall relative to the bounds.

### Actual Result

_No response_

8.2.7

5.0.0

8.1

iomedia

florianthomi:
The SQL selects products with inclusive bounds (price_min <= max), while the straddle test used a strict comparison (price_min < max). A product whose indexed price_min lands exactly on the requested upper bound therefore escaped the re-check and stayed in the listing at its real price, outside the range. The mirror case exists on the lower bound.

The extra clause is unnecessary: a re-check is needed exactly when the indexed envelope is not fully contained in the requested range, since a contained envelope guarantees the real price is in range. Also drop the int casts, which truncated non-integer bounds before comparison.

| Questions | Answers 
| Description?| A product whose indexed `price_min` lands exactly on the requested upper bound escapes the price re-check and stays in the listing at its real price, outside the range. The SQL selects with inclusive bounds while the straddle test uses a strict comparison.
| Type? | bug fix 
| BC breaks?| no
| Deprecations? | no
| Fixed ticket? | Fixes #1282
| How to test?| See below.
| Sponsor company | iomedia

### How to test

1. PrestaShop 8.x, `ps_facetedsearch` enabled with a price filter on a category.
2. Give the *Customer* group a discount (Customers > Groups) so products get a wide indexed range.
3. Rebuild the price index (module configuration, or `ps_facetedsearch-price-indexer.php`).
4. Pick a product, read its `price_min` in `ps_layered_price_index`, and browse the category **as a guest** with a price filter whose **upper bound equals that exact value**.

Before: the product is listed even though its real price is well above the bound. Any neighbouring bound excludes it correctly — only the one landing on `price_min` triggers it.

After: it is excluded, and it still shows up under the ranges that actually contain its price.

### DIFF
--- a/src/Filters/Products.php
+++ b/src/Filters/Products.php
@@ -170,8 +170,8 @@ class Products
     ) {
         /* for this case, price could be out of range, so we need to compute the real price */
         foreach ($matchingProductList as $key => $product) {
-            if (($product['price_min'] < (int) $priceFilter['min'] && $product['price_max'] > (int) $priceFilter['min'])
-                || ($product['price_max'] > (int) $priceFilter['max'] && $product['price_min'] < (int) $priceFilter['max'])
+            if ($product['price_min'] < $priceFilter['min']
+                || $product['price_max'] > $priceFilter['max']
             ) {
                 $price = Product::getPriceStatic($product['id_product'], $psLayeredFilterPriceUsetax);
                 if ($psLayeredFilterPriceRounding) {


ps-jarvis:
Hello @florianthomi!

Hlavtox (approved):

kpodemski:
> [!NOTE]

Thanks @florianthomi. I rebased your branch on current `dev` and pushed it, because the old CI runs were out of date. Tests pass: 116 tests, 1016 assertions.

**The bug is real and your fix works.**

I reproduced it with a product that costs 15.48 but has an indexed range of [12.00, 20.00], so `price_min` is exactly on the upper bound:

| filter | `dev` | your PR |
| --- | --- | --- |
| 10 to 12 | product is listed (wrong) | not listed (correct) |
| 10 to 11 | not listed | not listed |
| 10 to 20 | listed | listed |

Your analysis is correct. A re-check is needed when the indexed range is not fully inside the requested range.

**But it cannot be merged as it is.**

#1290 was merged after you opened this PR. It changed the SQL bounds so a product matches the whole range that the slider shows it in. A product at 15.48 is shown as covering 15 to 16, so it must match both 15 and 16.

Your change breaks that:

| Stationery, 3 products at 15.48, slider shows 15 to 16 | `dev` | your PR |
| --- | --- | --- |
| filter 15 to 15 | 3 products | 3 products |
| filter 16 to 16 | 3 products | **0 products** |

The reason: the SQL now accepts the product for filter 16, but your new re-check compares the real price 15.48 against the raw bound 16, sees it is lower, and removes it. This is the exact bug #1290 fixed, coming back on the upper end.

**Suggested change**

Compare using the same whole units the slider and the SQL use:

```php
$price = Product::getPriceStatic($product['id_product'], $psLayeredFilterPriceUsetax);

if (floor($price) > $priceFilter['max'] || ceil($price) < $priceFilter['min']) {
```

The `(int) $price` line has to go, because it removes the decimals that `ceil()` needs.

I tested this. It fixes #1282 and keeps #1290 working:

| filter | 10 to 12 | 10 to 11 | 15 to 15 | 16 to 16 | 10 to 20 |
| --- | --- | --- | --- | --- | --- |
| result | 0 | 0 | 3 | 3 | 3 |

All 116 tests still pass.

One side effect to be aware of: this makes `PS_LAYERED_FILTER_PRICE_ROUNDING` do nothing when "use tax" is on. In my sweep of 102 filter cases the flag changed 9 results today and 0 after this change. That is fine by me, and I will handle that setting separately. Worth knowing before you edit.

Happy to push the change to your branch if you prefer, just say so.

florianthomi:
Hi @kpodemski thanks for the review and the suggested changes. Feel free to push the changes straight to my branch :smiley:

PR #1289, opened 2026-08-09
Order a category listing by the position set in the browsed category

### LINKED ISSUE #26564: The position of the products by relevance is incorrect in the category page
### Prerequisites

- [X] I understand and accept the project's [code of conduct](https://github.com/PrestaShop/PrestaShop/blob/develop/CODE_OF_CONDUCT.md)

### Describe the bug and add screenshots

I have a problem with the position of products on the category page.

Here is the order of the products on category ID 9 as seen in the `ps_category_product` table:

image

Here is the order as seen in the back-office (identical to the one in the database):

image

And here is the order as seen on the category page (sorry, it's ugly):

image

The order does not follow the category order as you would expect.

The problem does not occur on all categories. It seems to come from the fact that we have a large number of products and categories and subcategories.

The problem has already been reported here but was closed: #20298.

On BO > Shop Parameters > Products, the default order is "Position inside category".

### Expected behavior

The position of the products is respected.

### Steps to reproduce

1. Create many categories and subcategories.
2. Create many products, each with multiple categories.
3. See the problem on some categories.

1.7.7.0

7.3.28

ps_facetedsearch (v3.7.1)

### LINKED ISSUE #42279: ps_facetedsearch: sorting a parent category by position uses the position from an arbitrary subcategory
### Describe the bug

When a category has subcategories, sorting its listing by position orders products by their position in an **arbitrary** category of the subtree instead of the one being browsed. Leaf categories are unaffected, so the shop looks fine everywhere except parent categories.

This is the same defect as #15062, #15278 and #21131. All three are closed as fixed, but the mechanism is still present in the current `dev` branch and in the module shipped with PrestaShop 8.2.7, so I am opening a fresh report rather than commenting on closed issues.

### Expected behaviour

A category listing sorted by position shows products in the order set for **that** category in the back office (`ps_category_product.position` for the browsed `id_category`).

### Actual behaviour

Real shop, category `Elektrické kolobežky` (id 19, four subcategories, 37 visible products, every product also assigned to a subcategory):

```
ps_category_product.position for id_category = 19 :  952, 953, 977, 978, 986, 984, 985, 857, …
what the storefront lists                         :  502, 317, 380, 392, 393, 394, 395, 409, …
```

The order shown is each product's position in *some* subcategory — 502 sits at position 0 in `Detské kolobežky`, 317 at position 0 in `Športové kolobežky`, and so on. Sorting by name or by price works correctly; only position is affected.

### Root cause

`Filters/Products.php` puts `position` into the initial population's select list, and `Adapter/MySQL.php` maps it to `category_product` with `joinCondition => '(p.id_product = cp.id_product)'` — no category restriction. `Product/Search.php` restricts the *rows* to the subtree with `nleft`/`nright`, not to the browsed category. The generated query is:

```sql
SELECT p.id_product FROM (
    SELECT p.id_product, …, cp.position
    FROM ps_product p
    INNER JOIN ps_category_product cp ON (p.id_product = cp.id_product)
    INNER JOIN ps_category c ON (cp.id_category = c.id_category AND c.active = 1)
    WHERE … AND c.nleft >= 5 AND c.nright <= 14
    GROUP BY p.id_product) p
INNER JOIN ps_category_product cp ON (p.id_product = cp.id_product)
GROUP BY p.id_product
ORDER BY p.position ASC, p.id_product DESC
```

`cp.position` is selected non-aggregated under `GROUP BY p.id_product`, so MySQL keeps the value from whichever row it happens to keep. A product assigned to both the parent and a child can therefore be ordered by either position, and which one wins is not defined.

PrestaShop sets `sql_mode = ''` in `DbPDO`, which is why this runs at all — under `ONLY_FULL_GROUP_BY` the query would be rejected outright instead of silently picking a row.

### Steps to reproduce

1. Take a category with at least one subcategory, and products assigned to both the parent and the child.
2. Give them a deliberate order in the parent category (Catalog → Categories → the category → drag the products).
3. Make sure `PS_LAYERED_FULL_TREE` is on (Faceted search → "Show products from subcategories").
4. Open the category on the front office sorted by position (the default, "Relevance").

The listing does not follow the order set in step 2. Repeat on a category with no subcategories and it does.

### A fix that works

Restricting the aggregate to the browsed category in `Filters/Products.php` is enough, and needs no change to the adapter because both `computeFieldName()` and `computeOrderByField()` pass any string containing `(` through untouched:

```php
if ($orderBy === 'position' && $query->getIdCategory()) {
    // makes the adapter join category_product in the outer query
    $this->searchAdapter->addSelectField('position');
    $orderBy = 'COALESCE(MIN(IF(cp.id_category = ' . (int) $query->getIdCategory()
        . ', cp.position, NULL)), 999999)';
}
$this->searchAdapter->setOrderField($orderBy);
```

`COALESCE` keeps products that exist only in a subcategory (they have no position in the parent) at the end rather than at the front, which is where a plain `MIN` would put the resulting `NULL`s.

Verified on the shop above across all 7 categories that have children and 3 that do not: every listing now matches `ps_category_product.position` for the browsed category, the per-category product counts are unchanged (37 / 230 / 120 — products from subcategories still appear), and the search page output is byte-identical to before the change. Happy to open a PR if the approach looks right.

### Additional information

- PrestaShop 8.2.7, ps_facetedsearch 5.0.0, PHP 8.3, MariaDB 10.11, multistore (4 shops).
- Also reproduced on the same shop before its migration: PrestaShop 1.7.6.2 with ps_facetedsearch 3.0.6 lists category 18 as `935,1079,685,1377,…` where the curated order is `1377,3,299,300,…`. The behaviour is long-standing, not a regression of 5.x.
- The `position` mapping in `Adapter/MySQL.php` on `dev` still carries `joinCondition => '(p.id_product = cp.id_product)'`.

boo-code:
| Questions         | Answers
| Description?      | Sorting a category listing by position orders products by their position in an arbitrary category of the subtree instead of the browsed one. A product has one position per category it belongs to; with `PS_LAYERED_FULL_TREE` the rows are restricted with `nleft`/`nright` rather than to the browsed category, so `cp.position` is not functionally dependent on the product the query groups by and the server keeps whichever row it likes. Leaf categories are unaffected, which is why a shop looks correct everywhere except parent categories. The ordering is now pinned to the browsed category, and a product reached only through a subcategory - which has no position there - is listed after the ones the merchant actually arranged.
| Type?             | bug fix
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | Fixes PrestaShop/PrestaShop#42279, Fixes PrestaShop/PrestaShop#26564.
| How to test?      | `composer test`. The added case asserts the generated SQL: the ordering expression is evaluated in the outer query, where `category_product` is joined, and is kept out of the initial population, which does not join it. Reverting only `src/Adapter/MySQL.php` makes the initial population select `ISNULL(MIN(IF(cp.id_category = 6, cp.position, NULL)))` against an alias it never joins, which is an `Unknown column 'cp.id_category'` at runtime. Manually: give products a deliberate order in a category that has subcategories and whose products also sit in those subcategories, turn on "Show products from subcategories", and browse it sorted by position.
| Sponsor company   |

### Why it is a real defect rather than a data quirk

The query is not valid aggregation - it only runs because PrestaShop sets `sql_mode = ''` in `DbPDO`.
Asking the same server to apply the standard rejects it outright:

```
ERROR 1055 (42000): Expression #2 of SELECT list is not in GROUP BY clause and contains
nonaggregated column 'prestashop.cp.position' which is not functionally dependent on columns
in GROUP BY clause; this is incompatible with sql_mode=only_full_group_by
```

So which position wins is undefined, not merely unexpected. On the default data set of a 9.1 shop the
server happens to keep the browsed category's row and the listing looks right; forcing the other
candidate rows on the same data gives `16,6,17,7,18,8,…` where the merchant's order is
`7,6,8,10,9,11,…`. That is the same symptom #42279 reports from a production shop, so this fixes a
latent defect whose visibility depends on the plan and the data rather than one that reproduces
everywhere.

### The two changes

`src/Filters/Products.php` pins the ordering:

```php
$positionInCategory = 'MIN(IF(cp.id_category = ' . (int) $query->getIdCategory() . ', cp.position, NULL))';
$orderBy = 'ISNULL(' . $positionInCategory . ') ASC, ' . $positionInCategory;
```

`ISNULL(...) ASC` keeps the products that have no position in the browsed category last in both
directions, which a sentinel value would not do on `DESC`.

`src/Adapter/MySQL.php` stops pushing an expression order field into the initial population.
`computeOrderByField()` already treats a value containing `(` as an outer-query expression and returns
it untouched, but it had added it to the initial population's select list first. The initial population
does not carry the outer query's joins, so the expression there refers to tables that are not joined at
that level. Any caller passing an expression hits this, not just this one.

### DIFF
--- a/src/Adapter/MySQL.php
+++ b/src/Adapter/MySQL.php
@@ -444,8 +444,11 @@ class MySQL extends AbstractAdapter
         }
 
         // Expose sortable fields from the initial population when their mapped value has a stable output name.
+        // An expression is skipped: it is evaluated against the outer query's joins, which the initial population
+        // does not have, so selecting it there refers to tables that are not joined at that level.
         if ($this->getInitialPopulation() !== null
             && $orderField !== 'price'
+            && strpos($orderField, '(') === false
             && (
                 !isset($filterToTableMapping[$orderField]['fieldName'])
                 || isset($filterToTableMapping[$orderField]['fieldAlias'])
--- a/src/Filters/Products.php
+++ b/src/Filters/Products.php
@@ -71,6 +71,18 @@ class Products
         $orderWay = Validate::isOrderWay($orderWay) ? $orderWay : 'ASC';
         $orderBy = Validate::isOrderBy($orderBy) ? $orderBy : 'position';
 
+        // A product has one position per category it belongs to. When the listing spans a subtree
+        // (PS_LAYERED_FULL_TREE), the rows are restricted with nleft/nright rather than to the browsed
+        // category, so `position` is not functionally dependent on the product the query groups by and
+        // the value picked is whichever row the server happens to keep. Pin the ordering to the browsed
+        // category; a product reached only through a subcategory has no position here and is listed after
+        // the ones the merchant did arrange.
+        if ($orderBy === 'position' && $query->getIdCategory()) {
+            $positionInCategory = 'MIN(IF(cp.id_category = ' . (int) $query->getIdCategory() . ', cp.position, NULL))';
+            $this->searchAdapter->addSelectField('position');
+            $orderBy = 'ISNULL(' . $positionInCategory . ') ASC, ' . $positionInCategory;
+        }
+
         // Apply it to the filter
         $this->searchAdapter->setOrderField($orderBy);
         $this->searchAdapter->setOrderDirection($orderWay);
--- a/tests/php/FacetedSearch/Adapter/MySQLTest.php
+++ b/tests/php/FacetedSearch/Adapter/MySQLTest.php
@@ -70,6 +70,30 @@ class MySQLTest extends MockeryTestCase
         Configuration::setStaticExpectations($configurationMock);
     }
 
+    /**
+     * A product has one position per category, so ordering a subtree listing by position has to
+     * aggregate over the browsed category. The expression is evaluated against the outer query's
+     * joins, so it must not be selected in the initial population, which does not have them.
+     */
+    public function testGetQueryOrderedByAnExpressionKeepsItOutOfTheInitialPopulation()
+    {
+        $this->adapter->useFiltersAsInitialPopulation();
+        $this->adapter->addFilter('nleft', [9], '>=');
+        $this->adapter->addFilter('nright', [14], '<=');
+        $this->adapter->addGroupBy('id_product');
+        $this->adapter->addSelectField('position');
+        $this->adapter->setOrderField('ISNULL(MIN(IF(cp.id_category = 6, cp.position, NULL))) ASC, MIN(IF(cp.id_category = 6, cp.position, NULL))');
+        $this->adapter->setOrderDirection('asc');
+
+        $query = $this->adapter->getQuery();
+        $innerQuery = substr($query, strpos($query, '(') + 1, strrpos($query, ') p INNER JOIN') - strpos($query, '(') - 1);
+
+        $this->assertNotContains('cp.id_category', $innerQuery, 'The initial population does not join category_product, so it cannot select an expression using it.');
+        $this->assertNotContains('ps_category_product', $innerQuery, 'The expression is read in the outer query, so the initial population does not need the join at all.');
+        $this->assertContains(') p INNER JOIN ps_category_product cp ON (p.id_product = cp.id_product)', $query, 'The outer query joins category_product, which is what the expression reads.');
+        $this->assertContains('ORDER BY ISNULL(MIN(IF(cp.id_category = 6, cp.position, NULL))) ASC, MIN(IF(cp.id_category = 6, cp.position, NULL)) ASC, p.id_product DESC', $query);
+    }
+
     public function testGetEmptyQuery()
     {
         $this->assertEquals(


Touxten:
Close and open for news tests

kpodemski (approved):
Rebased on `dev` and force-pushed — the conflict was in `computeOrderByField()`, where cae0c9c had rewritten the same condition. Both guards are needed, so they are now combined:

```php
if ($this->getInitialPopulation() !== null
    && $orderField !== 'price'
    && strpos($orderField, '(') === false
    && (
        !isset($filterToTableMapping[$orderField]['fieldName'])
        || isset($filterToTableMapping[$orderField]['fieldAlias'])
    )
) {
```

`dev`'s condition alone does not cover this case: an expression is not a key in the mapping, so `!isset(...['fieldName'])` is true and it would still be pushed into the initial population.

All CI green. Locally 108 tests / 976 assertions pass, and reverting only `src/Adapter/MySQL.php` turns `testGetQueryOrderedByAnExpressionKeepsItOutOfTheInitialPopulation` red.

### On "undefined, not merely unexpected"

Worth confirming this out loud, because the listing looks **correct on default 9.2 data** — I get the merchant's order `7,6,8,10,9,11,15,16,17,18,19` for *Accessories* both with and without the patch. The defect is latent, exactly as the description says. What is demonstrable on that same data is that the old query is not valid aggregation:

```
sql_mode as PrestaShop connects: ''

OLD (selects bare cp.position) -> REJECTED: 'prestashop92.cp.position' isn't in GROUP BY
NEW (pinned to category 6)     -> accepted: 7,6,8,10,9,11,15,16,17,18,19
```

And an equally legitimate choice of row on the same data gives a different listing, which is the point:

```
ORDER BY cp.position      -> 7,6,8,10,9,11,15,16,17,18,19
ORDER BY MIN(cp.position) -> 7,6,16,17,8,18,10,9,11,15,19
```

### ISNULL branch

Demo data has no product that reaches a category only through a subcategory, so I forced it by pinning to category 7 while browsing the subtree of 6, where only 16/17/18 have a position:

```
ASC  -> 16,17,18, 19,7,6,8,10,9,11,15
DESC -> 18,17,16, 19,7,6,8,10,9,11,15
```

Unpositioned products stay last in both directions, which is the behaviour a sentinel value would not give.

### Other checks

- Leaf categories unchanged (*Stationery* `16,17,18` / `18,17,16`), parent *Home* follows its own positions, search results unaffected.
- Guarded by `$query->getIdCategory()`, so search, manufacturer and supplier listings keep the old path.
- `addSelectField('position')` is what forces the `category_product` join into the outer query that the expression reads; every product in a category listing has a row there, and the `GROUP BY id_product` collapses the duplicates, so nothing is dropped or repeated — 11 distinct products before and after.

One gap, not a blocker: the assertion covers `MySQL.php`, but the ordering expression built in `Filters/Products.php` has no test of its own, so a future edit to that string would not be caught.

LGTM.

PR #1290, opened 2026-08-10
Match the price filter to the range the slider displays

### LINKED ISSUE #37604: Basic price filtering of faceted search is broken
### Prerequisites

- [x] I have already [searched in existing issues](https://github.com/PrestaShop/PrestaShop/issues?q=is%3Aissue+label%3ABug) and found no previous report of this bug.

I have following product. There is `5995,000001` because there is no other way to save that price.

Image

🔴 **Issue 1 - I go to FO, and I see available price filter 5995 -> 5996.**

Image

🔴**Issue 2 - I filter by exact price 5995 and no products are displayed.**

Image

**Now when I modify the price tax excluded a bit so it's slightly lower than 5995.**

Image

🔴 The 5996 number disappears, but the filter still doesn't work because the product price is `5994,999934`.

Image

### Expected behavior

Basic things like this should work.

### Steps to reproduce

See issue description

8.1.7

8.1

ps_facetedsearch v3.16.1

_No response_

boo-code:
| Questions         | Answers
| Description?      | The price slider is built from `floor()` of the lowest indexed price and `ceil()` of the highest, while the filter compares the raw indexed column against the value the shopper picked. A product indexed at 5995.000001 is therefore offered as spanning 5995 to 5996 and then satisfies neither end, so dragging either handle empties the listing. The bounds now cover the range that was displayed for the product.
| Type?             | bug fix
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | Fixes PrestaShop/PrestaShop#37604.
| Sponsor company   |

### How to test

Give a product a tax excluded price whose tax included value is not representable exactly, for instance 4995.833334 at 20 percent, and reindex. The shop then holds a single indexed value of 5995.000001.

Measured on a 9.2 shop before the change:

```
indexed: price_min=5995.000001 price_max=5995.000001
slider shows: 5995 -> 5996

filter 5995..5995 -> product returned: no
filter 5996..5996 -> product returned: no
filter 5995..5996 -> product returned: YES
```

Both ends of the range the slider itself offered return nothing. That is the two symptoms of the report, the unexpected upper bound and the empty result on an exact price.

### The change

```php
$this->getSearchAdapter()->addFilter('price_min', [floor($maxPrice) + 1], '<');
$this->getSearchAdapter()->addFilter('price_max', [ceil($minPrice) - 1], '>');
```

`price_min < floor(max) + 1` is the same condition as `floor(price_min) <= max`, and the mirror on the other side, so a product matches exactly the range it was shown as occupying. Writing it this way keeps the column alone on its side of the comparison, so the index on `layered_price_index` is still usable, which `FLOOR(price_min) <= :max` would have prevented.

The six decimals themselves are not the problem, they come from `ps_round($price * (100 + $taxRate) / 100, 6)` at index time and are correct. Only the comparison was using a different rounding from the display.

### Tests

`testInitSearchWidensPriceBoundsToTheDisplayedRange` pins the produced bounds for a 5995 to 5995 selection, and the existing expectation in `testInitSearchWithAllFilters` is updated to the new operators. Reverting the change turns both red.

```
OK (101 tests, 938 assertions)
```

### Related

#1281 and #1283 also touch price filtering, both in the post filter of `Filters/Products.php`. This one changes the bounds sent to SQL in `Product/Search.php`, so they do not overlap, but they are worth sequencing deliberately since all three affect which products a price range returns.

### DIFF
--- a/src/Product/Search.php
+++ b/src/Product/Search.php
@@ -493,7 +493,12 @@ class Search
      */
     private function addPriceFilter($minPrice, $maxPrice)
     {
-        $this->getSearchAdapter()->addFilter('price_min', [$maxPrice], '<=');
-        $this->getSearchAdapter()->addFilter('price_max', [$minPrice], '>=');
+        // The slider is built from floor() of the lowest indexed price and ceil() of the highest,
+        // so a product indexed at 5995.000001 is offered as spanning 5995 to 5996. Comparing the
+        // raw column against those bounds then rejects it at both ends of its own range.
+        // `price_min < floor(max) + 1` is the same condition as `floor(price_min) <= max`, and
+        // keeps the column alone on its side of the comparison so the index is still usable.
+        $this->getSearchAdapter()->addFilter('price_min', [floor($maxPrice) + 1], '<');
+        $this->getSearchAdapter()->addFilter('price_max', [ceil($minPrice) - 1], '>');
     }
 }
--- a/tests/php/FacetedSearch/Product/SearchTest.php
+++ b/tests/php/FacetedSearch/Product/SearchTest.php
@@ -334,16 +334,18 @@ class SearchTest extends MockeryTestCase
                     ],
                 ],
                 'price_min' => [
-                    '<=' => [
+                    // The bounds are widened by one unit so a product matches the range the slider
+                    // displayed for it, which is floor(price_min) to ceil(price_max).
+                    '<' => [
                         [
-                            200.0,
+                            201.0,
                         ],
                     ],
                 ],
                 'price_max' => [
-                    '>=' => [
+                    '>' => [
                         [
-                            50.0,
+                            49.0,
                         ],
                     ],
                 ],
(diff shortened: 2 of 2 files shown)


Touxten:
Close and open for news tests

Hlavtox (approved):

kpodemski (approved):
Tested on PS 9.2. Confirmed, and the default demo data already reproduces it — no crafted price needed.

Category **Stationery** holds 3 products at €15.48, so the slider offers €15.00 – €16.00:

| filter | before | after |
| --- | --- | --- |
| 15 – 15 | 0 | 3 |
| 16 – 16 | 0 | 3 |
| 15 – 16 | 3 | 3 |
| 14 – 14 | 0 | 0 |
| 17 – 17 | 0 | 0 |

Both ends of the slider's own range returned nothing, exactly as described:

price-before-empty

price-after-3-products

14 and 17 still return nothing, so the widening stays inside the range the product was displayed as occupying rather than leaking a whole unit outwards. In *Home*, 34 – 34 goes 0 → 4: the €34.80 products were previously unreachable at their own displayed value.

Checks I made beyond the description:

- `getMinMaxPriceValue()` really does return `[floor(min), ceil(max)]`, so the premise holds.
- The algebra is right: for an integer bound, `price_min < floor(max) + 1` ⟺ `floor(price_min) <= max`, and the mirror on the other side. Column stays bare, so the `layered_price_index` index is still usable.
- Scoped to price only — the weight slider uses raw floats in `getMinMaxValue()` and is unaffected.
- The `pricePostFiltering()` in `Filters/Products.php` does not undo it. It only runs with `PS_LAYERED_FILTER_PRICE_USETAX`/`_ROUNDING` on (both were on here) and only fires for products straddling a bound, so single-price products newly admitted are not dropped again.
- 101 tests / 938 assertions pass; reverting `src/Product/Search.php` turns exactly the two named tests red.

One thing worth stating so nobody reads it as a regression: with rounding on, some ranges show a product count higher than the number of tiles rendered (`25 – 40` gives 5 vs 4 in *Home*). That is the post filter removing a product after it was counted, and it reproduces identically on `dev` without this PR — the same area #1281 and #1283 are in.

LGTM.

PR #1292, opened 2026-08-17
Support combination feature values in feature filters

jf-viguier:
| Questions         | Answers
| Description?      | See below.
| Type?             | new feature
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | https://github.com/PrestaShop/PrestaShop/discussions/37479
| Sponsor company   | creabilis.com

### Description

PrestaShop 9.3 introduced feature values at combination level ([PrestaShop/PrestaShop#41830](https://github.com/PrestaShop/PrestaShop/pull/41830)). This PR brings them into the faceted search: feature filters now also take combination (`product_attribute`) feature values into account, so a feature carried only by combinations shows up in the facets with correct product counts.

The behavior is gated first by PrestaShop `>= 9.3`, then by the `combination_feature_values` feature flag. On older versions, or with the flag off, nothing changes.

### About `rawTable`

The MySQL adapter maps each filter field to a DB table, and the query builder prepends `_DB_PREFIX_` to that name. To merge product-level and combination-level values, `id_feature` / `id_feature_value` are now mapped to a derived sub-select (a `UNION` of `feature_product` and combination values resolved through `product_attribute`) instead of `feature_product`. Since a derived table is already a full SQL expression and not a table name, a new `rawTable` flag on the join mapping tells the builder to use it as-is, without prepending the prefix.

### How to test

1. On PrestaShop `>= 9.3`, enable the `combination_feature_values` feature flag.
2. Create a feature and assign its values only to a product's combinations (not to the product itself).
3. Add that feature to the filter template of the product's category.
4. Front office: the feature appears in the faceted filters, with the combination values and correct product counts.

### DIFF
--- a/src/Adapter/MySQL.php
+++ b/src/Adapter/MySQL.php
@@ -176,7 +176,18 @@ class MySQL extends AbstractAdapter
         $featureJoinCondition = '(p.id_product = fp.id_product)';
         $featureJoinExtra = [];
         if ($this->isCombinationFeatureFilteringEnabled()) {
-            $featureProductTable = $this->getFeatureProductDerivedTable();
+            // Derived table (id_product, id_product_attribute, id_feature, id_feature_value) merging
+            // product-level feature values with the ones defined at combination level
+            // (feature_product_attribute, resolved to their product through product_attribute). The
+            // id_product_attribute column is NULL for product-level values, so a feature filter can be
+            // correlated with a specific combination. The UNION removes duplicates so a value defined
+            // at both levels is not counted twice.
+            $featureProductTable = '(SELECT id_product, NULL AS id_product_attribute, id_feature, id_feature_value'
+                . ' FROM ' . _DB_PREFIX_ . 'feature_product'
+                . ' UNION'
+                . ' SELECT pa.id_product, pa.id_product_attribute, fpa.id_feature, fpa.id_feature_value'
+                . ' FROM ' . _DB_PREFIX_ . 'feature_product_attribute fpa'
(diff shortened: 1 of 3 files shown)


Touxten (dismissed):

src/Adapter/MySQL.php
```diff
@@ -339,6 +367,41 @@ protected function getFieldMapping()
...
+     *
+     * Overrides must keep the exact same columns (id_product, id_product_attribute, id_feature,
+     * id_feature_value) so the derived table stays a drop-in replacement for feature_product in
+     * getFieldMapping().
+     *
+     * @return string
+     */
+    protected function getFeatureProductDerivedTable()
+    {
+        return '(SELECT id_product, NULL AS id_product_attribute, id_feature, id_feature_value'
```
Hlavtox: This can just be inlined

Touxten:
Close and open for news tests

src/Adapter/MySQL.php
jf-viguier: @Hlavtox thanks, I've inlined, indeed this class cannot be overrided,no need to be a separated method.

Touxten (approved):

kpodemski (approved):

Touxten:
Ok for me Thank you @jf-viguier 


https://github.com/user-attachments/assets/cc6e3a04-c9a3-4068-b467-a4a73337bf4e

jf-viguier:
Thanks @kpodemski @Touxten 👍

PR #1303, opened 2026-08-27
Optimize performance of product listing

Hlavtox:
| Questions         | Answers
| Description?      | Optimizes the SQL query of the module - the initial population of products to only add fields if they are required. It reduced the load time in profiler by average of 40 ms on all of my listings, you can try and you should get same results. Next, fixes a small issue where DataAcccessor cached values were wiped, but this did not bring measurable performance benefit, maybe it would be visible in some cases I did not try.
| Type?             | refacto
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | 
| How to test?      | 

### Performance measurement on combination heavy page

| Load | Default      | PR 1303      |
|------|-------------:|-------------:|
| 1    | 398 ms       | 220 ms       |
| 2    | 395 ms       | 239 ms       |
| 3    | 434 ms       | 220 ms       |
| 4    | 388 ms       | 228 ms       |
| 5    | 376 ms       | 226 ms       |
| 6    | 383 ms       | 217 ms       |
| 7    | 424 ms       | 221 ms       |
| 8    | 387 ms       | 224 ms       |
| 9    | 397 ms       | 221 ms       |
| 10   | 404 ms       | 222 ms       |
| **Average** | **398,6 ms** | **223,8 ms** |
| **Rows**    | **55 923 840** | **1 997 280** |

### DIFF
--- a/src/Adapter/MySQL.php
+++ b/src/Adapter/MySQL.php
@@ -556,6 +556,28 @@ class MySQL extends AbstractAdapter
     }
 
     /**
+     * Check whether a field must be read from its mapped table when used in a filter condition.
+     *
+     * Aggregated fields are exposed by the derived product table as an aggregate
+     * (SUM(sa.quantity) as quantity), which is not interchangeable with the row level value a
+     * WHERE condition compares against. They always have to be read from their mapped table,
+     * otherwise the filter silently changes meaning.
+     *
+     * @param string $fieldName
+     * @param array $filterToTableMapping
+     *
+     * @return bool
+     */
+    private function requiresMappedTableForFilter($fieldName, array $filterToTableMapping)
+    {
+        if (isset($filterToTableMapping[$fieldName]['aggregateFunction'])) {
+            return true;
+        }
+
+        return $this->requiresMappedTable($fieldName, $filterToTableMapping);
+    }
+
+    /**
      * Add alias to table field name
      *
      * @param string $fieldName
@@ -620,7 +642,7 @@ class MySQL extends AbstractAdapter
                 foreach ($operations as $idx => $operation) {
                     $selectAlias = 'p';
                     $values = $operation[1];
-                    if ($this->requiresMappedTable($operation[0], $filterToTableMapping)) {
+                    if ($this->requiresMappedTableForFilter($operation[0], $filterToTableMapping)) {
                         $joinMapping = $filterToTableMapping[$operation[0]];
                         // If index is not the first, append to the table alias for
                         // multi join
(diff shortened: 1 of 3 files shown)


kpodemski (changes requested):
Hello Daniel

40ms isn't a difference I'd aim, if the alternative is not to have some data at all time. I've checked this PR only on mobile and will take a closer look later, but that's how I understood it.

There are PR in the review queue optimizing this module in seconds per request, and I think we should focus on that, then come back to this one 👍🏻

Hlavtox:
@kpodemski I added the table with timings. :-)

kpodemski:
Benchmarked this on a large catalogue and the numbers are much better than the ~40 ms you mentioned. Setup: PS 9.1.5 / PHP 8.3 / MariaDB 11.4, 30 000 products, 120 000 combinations, 240 000 feature-product rows, 150 000 stock rows, availability + price + 8 feature + 3 attribute group filters enabled. Baseline is this PR's merge base (`580ac6e`).

| | base | this PR |
|---|---|---|
| FO category page, 1 250 products | 1.20 s | **0.54 s** |
| FO category page, 30 000 products | 24.7 s | **5.3 s** |
| Facet block, 30 000-product category | 23.1 s | **4.58 s** |
| Total SQL time across the benchmark | 60.2 s | **13.4 s** |

Same 464 queries either way — `product_sale` joins drop 432 → 0 and `stock_available` 436 → 100.

While verifying I found one behaviour change, so I pushed a fix straight onto the branch (cae0c9c) — shout if you'd rather I had opened it against your fork instead.

**Regression:** with `PS_LAYERED_FILTER_SHOW_OUT_OF_STOCK_LAST` enabled, the "Not available" availability facet reported 0 and disappeared from the page, while `?q=Availability-Not+available` still returned 200 products.

`computeShowLast()` adds `quantity` to the initial population, where it lives as `SUM(sa.quantity)`. After that `requiresMappedTable()` sees `quantity` as already exposed by the derived table, so the facet's filter was built against the aggregate:

```sql
base:     WHERE sa.quantity <= 0 AND sa_1.out_of_stock IN (0,2)   -> 200
this PR:  WHERE p.quantity  <= 0 AND sa_1.out_of_stock IN (0,2)   -> 0
```

The product list applies the same filter *inside* the initial population on row-level `sa.quantity`, so it kept the old meaning — facet and filter disagreed.

**Fix:** a `requiresMappedTableForFilter()` variant that always reads aggregate-mapped fields from their mapped table in `computeWhereConditions()` and the joins backing them. `computeFieldName()` / ORDER BY are untouched, so "out of stock last" still sorts on the aggregate exactly as before. `quantity` is the only mapping with `aggregateFunction`, so the blast radius is small. Added a regression test (it fails on the previous commit with `WHERE ((p.quantity<=0))`).

Re-verified over 176 cases (2 categories × 8 filter sets × 8 sort orders × out-of-stock-last on/off) plus 15 real front-office URLs covering price sorting, the price slider, availability, feature and attribute facets: **PR + fix is identical to base on product counts, product order and every facet count**, and performance is unchanged (13.8 s vs 13.4 s SQL — within noise). Your existing test expectations pass untouched, and `composer run test` is green (102 tests).

kpodemski (approved):

kpodemski:
Info to everyone: I'm merging this one. I've tested it manually as well, we'll perform another QA after 1304

ps-jarvis:
PR merged, well done!

Message to @PrestaShop/committers: do not forget to milestone it before the merge.

PR #1304, opened 2026-08-27
Add option for AND checkboxes to feature filter

Hlavtox:
| Questions         | Answers
| Description?      | Feature filter used to work with joining X of feature tables over and over again. Refactored to a simple exists. Also, it allowed me to implement AND option for checkboxes. So, now you can easily setup filters with "Required features" of a product, like connectors on a PC for example.
| Type?             | refacto
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | 
| How to test?      | 

## Performance comparsion on Shop 1 - [trendo.cz](https://www.trendo.cz/117-zvyhodnene-sety?q=Materi%C3%A1l+d%C5%99ezu-Granit/Typ+materi%C3%A1lu-Granital/Tvar+d%C5%99ezu-D%C5%99ezy+s+odkapem/Zp%C5%AFsob+mont%C3%A1%C5%BEe-Horn%C3%AD+%5C-+na+pracovn%C3%AD+desku/Typ+baterie-Granitov%C3%A1+s+vytahovac%C3%AD+koncovkou/P%C5%99esn%C3%BD+rozm%C4%9Br+%5C-+d%C3%A9lka+%28mm%29-780)

It takes about 400ms on profiling, 100 ms to execute the select query, [with the previous PR](https://github.com/PrestaShop/ps_facetedsearch/pull/1303).

With this PR on top, it takes about 350ms on profiling, 60 ms to execute the select query.

## Performance comparsion on shop 2 - [lacne-elektromotory.sk](https://www.lacne-elektromotory.sk/63-elektromotory?q=Nap%C3%A4tie+%28V%29-230+V/V%C3%BDkon+%28kW%29-0%2C12+kW-0%2C13+kW-0%2C18+kW-0%2C25+kW-0%2C29+kW-0%2C31+kW-0%2C37+kW-0%2C55+kW-0%2C75+kW-1%2C1+kW-1%2C5+kW-3+kW/Kostra-Hlin%C3%ADk/Vinutie+elektromotora-Meden%C3%A9)

| Load | PR 1303      | PR 1304      |
|------|-------------:|-------------:|
| 1    | 208 ms       | 177 ms       |
| 2    | 178 ms       | 170 ms       |
| 3    | 177 ms       | 175 ms       |
| 4    | 185 ms       | 179 ms       |
| 5    | 179 ms       | 176 ms       |
| 6    | 183 ms       | 182 ms       |
| 7    | 179 ms       | 175 ms       |
| 8    | 181 ms       | 173 ms       |
| 9    | 180 ms       | 181 ms       |
| 10   | 177 ms       | 192 ms       |
| **Average** | **182,7 ms** | **178 ms** |
| **Rows**    | **47 615 946** | **47 615 946** |

I would not call it a measurable performance benefit. SQL engine probably optimizes it most of the time.

### Screenshot
Snímek obrazovky 2026-08-28 005650

### DIFF
--- a/src/Adapter/MySQL.php
+++ b/src/Adapter/MySQL.php
@@ -426,6 +426,34 @@ class MySQL extends AbstractAdapter
     }
 
     /**
+     * Build a feature value condition without multiplying rows in the product query.
+     *
+     * @param array $values
+     *
+     * @return string
+     */
+    private function computeFeatureValueExistsCondition(array $values)
+    {
+        // Use equality for one value and IN for multiple OR values.
+        if (count($values) === 1) {
+            $valueCondition = '=' . $this->getJoinedEscapedValue(', ', $values);
+        } else {
+            $valueCondition = ' IN (' . $this->getJoinedEscapedValue(', ', $values) . ')';
+        }
+
+        // Match values assigned directly to the product.
+        $productFeatureCondition = 'EXISTS (SELECT 1 FROM ' . _DB_PREFIX_ . 'feature_product fp_filter WHERE fp_filter.id_product = p.id_product AND fp_filter.id_feature_value' . $valueCondition . ')';
+        if (!$this->isCombinationFeatureFilteringEnabled()) {
+            return $productFeatureCondition;
+        }
+
+        // Match combination values against the same combination row used by attribute filters.
+        $combinationFeatureCondition = 'EXISTS (SELECT 1 FROM ' . _DB_PREFIX_ . 'feature_product_attribute fpa_filter WHERE fpa_filter.id_product_attribute = pa.id_product_attribute AND fpa_filter.id_feature_value' . $valueCondition . ')';
+
+        return '(' . $productFeatureCondition . ' OR ' . $combinationFeatureCondition . ')';
+    }
+
+    /**
      * Compute the orderby fields, adding the proper alias that will be added to the final query
      *
      * @param array $filterToTableMapping
@@ -702,6 +730,12 @@ class MySQL extends AbstractAdapter
             foreach ($filterOperations as $operations) {
                 $conditions = [];
                 foreach ($operations as $idx => $operation) {
+                    // Filter feature values through EXISTS so the main query keeps one row per current relation.
+                    if ($operation[0] === 'id_feature_value' && (empty($operation[2]) || $operation[2] === '=')) {
+                        $conditions[] = $this->computeFeatureValueExistsCondition($operation[1]);
+                        continue;
+                    }
+
                     $selectAlias = 'p';
                     $values = $operation[1];
                     if ($this->requiresMappedTableForFilter($operation[0], $filterToTableMapping)) {
@@ -818,6 +852,15 @@ class MySQL extends AbstractAdapter
         foreach ($this->getOperationsFilters() as $filterOperations) {
             foreach ($filterOperations as $operations) {
                 foreach ($operations as $idx => $operation) {
+                    // Feature value operations use EXISTS and only combination values require the shared combination join.
+                    if ($operation[0] === 'id_feature_value' && (empty($operation[2]) || $operation[2] === '=')) {
+                        if ($this->isCombinationFeatureFilteringEnabled()) {
+                            $this->addJoinConditions($joinList, $filterToTableMapping['id_product_attribute'], $filterToTableMapping);
+                        }
+
+                        continue;
+                    }
+
                     if ($this->requiresMappedTableForFilter($operation[0], $filterToTableMapping)) {
                         $joinMapping = $filterToTableMapping[$operation[0]];
                         if ($idx !== 0 || $operationIdx !== 0) {
@@ -1020,6 +1063,21 @@ class MySQL extends AbstractAdapter
 
         $operationsFilters = clone $initialPopulation->getOperationsFilters();
         foreach ($operationsFilters as $operationName => $operations) {
+            // Feature EXISTS filters already restrict product membership in the initial population.
+            $containsOnlyFeatureValueOperations = true;
+            foreach ($operations as $operationGroup) {
+                foreach ($operationGroup as $operation) {
+                    if ($operation[0] !== 'id_feature_value') {
+                        $containsOnlyFeatureValueOperations = false;
+                        break 2;
+                    }
+                }
+            }
+
+            if ($containsOnlyFeatureValueOperations) {
+                continue;
+            }
+
             $this->addOperationsFilter(
                 $operationName,
                 $operations
--- a/src/Adapter/MySQL.php
+++ b/src/Adapter/MySQL.php
@@ -426,34 +426,6 @@ class MySQL extends AbstractAdapter
     }
 
     /**
-     * Build a feature value condition without multiplying rows in the product query.
-     *
-     * @param array $values
-     *
-     * @return string
-     */
-    private function computeFeatureValueExistsCondition(array $values)
-    {
-        // Use equality for one value and IN for multiple OR values.
-        if (count($values) === 1) {
-            $valueCondition = '=' . $this->getJoinedEscapedValue(', ', $values);
-        } else {
-            $valueCondition = ' IN (' . $this->getJoinedEscapedValue(', ', $values) . ')';
-        }
-
-        // Match values assigned directly to the product.
-        $productFeatureCondition = 'EXISTS (SELECT 1 FROM ' . _DB_PREFIX_ . 'feature_product fp_filter WHERE fp_filter.id_product = p.id_product AND fp_filter.id_feature_value' . $valueCondition . ')';
-        if (!$this->isCombinationFeatureFilteringEnabled()) {
-            return $productFeatureCondition;
-        }
-
-        // Match combination values against the same combination row used by attribute filters.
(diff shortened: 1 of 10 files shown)


kpodemski:
I'm skeptical about the performance-related PRs you sent us, Daniel. Why? Because the machine you host your store on isn't easy to compare with others. Differences of around 20-80ms can be due to the configuration and specifics of your environment. I remember when you saw differences for `ProductLazyArray` whereas everyone else saw a slower query.

That being said, it's a bit painful to test, but we'd need tests from others on different configurations using a different dataset.

ps-jarvis:

    - [ ] `Checkbox (AND)`

kpodemski:
Ran this through the same bench I used on #1303 — PS 9.1.5 / PHP 8.3 / MariaDB 11.4, 30 000 products, 120 000 combinations, 240 000 feature-product rows, availability + price + 8 feature + 3 attribute group filters. Baseline is current `dev` (`12a26f4`, so #1303 included). Two rounds with the order reversed, agreeing to ~1%.

**The AND feature works exactly as advertised.** On data where products carry several values of one feature, `?q=Bench+Feature+1-F1\-V6-F1\-V11` gives 11 products with `filter_type=4` and 233 with `filter_type=0` — both matching SQL ground truth. `dev` degrades gracefully to OR for the unknown type. No correctness regressions either: a 176-case matrix (2 categories × 8 filter sets × 8 sort orders × out-of-stock-last on/off) is identical to `dev` on product counts, product order and every facet count.

**But the `EXISTS` refactor costs on my dataset**, total benchmark SQL 13.70 s → 15.43 s (+12.6%):

| scenario | dev | this PR |
|---|---|---|
| 1 feature selected | 0.553 s | **0.495 s** |
| 2 features | 0.132 s | 0.294 s (2.2×) |
| 3 features | 0.103 s | 0.294 s (2.9×) |
| 2 features + colour | 0.179 s | 0.553 s (3.1×) |
| 2 features + in stock | 0.802 s | 1.035 s (1.3×) |
| no filter (30 000 products) | 4.66 s | 4.60 s |

The product-list query is unaffected (10.4 vs 10.5 ms) and the `EXISTS` rewrite itself is neutral — the population subquery measures 11.5 ms as a join and 11.2 ms as `EXISTS`, same plan. The cost is in the **facet counts**: on `dev` they run in 3–4 ms each, here they all converge on ~11 ms, i.e. the full population cost. EXPLAIN of one of them:

```
dev :  fp_2        ref  id_feature_value  rows=1495
       <derived2>  ref  key0              rows=1       -> population evaluated per surviving product
       4.2 ms

PR  :  <derived2>  ALL  (no key)          rows=10465   -> population fully materialised, then scanned
       fp_filter   ref  id_feature_value  rows=1495  Start temporary
       11.9 ms
```

So the outer `INNER JOIN ps_feature_product fp … WHERE fp.id_feature_value = X` — the "redundant" condition that `copyOperationsFilters()` now skips — is exactly what lets MariaDB split-materialise the derived table. Without it every facet query re-evaluates the whole population.

I tried three ways to keep the `EXISTS` and get that back:

| variant | 1 feat | 2 feat | 3 feat | 2 feat+colour | 2 feat+stock |
|---|---|---|---|---|---|
| dev | 0.553 | 0.132 | 0.103 | 0.179 | 0.802 |
| this PR | **0.495** | 0.294 | 0.294 | 0.553 | 1.035 |
| V1 — drop the `copyOperationsFilters` skip, keep outer conditions | 0.502 | 0.343 | 0.298 | **0.196** | **0.528** |
| V2 — keep the AND feature, revert `MySQL.php` to dev | 0.559 | **0.132** | **0.102** | 0.188 | 0.785 |
| V3 — `EXISTS` only in the innermost population, joins outside | 0.551 | 0.336 | 0.302 | 0.322 | 0.565 |

An outer `EXISTS` doesn't restore the pushdown (V1), and neither does confining `EXISTS` to the population (V3). **V2 — keeping Converter/Search/Block and reverting `src/Adapter/MySQL.php` — is the only variant with no regression**, identical to dev on all 176 cases and every scenario, with the AND filter still working. It's a straight revert of that one file; happy to push it if you want it.

Worth saying plainly: this is one dataset on one engine, and your shops clearly behaved differently. My catalogue has 8 features per product, so the row multiplication `EXISTS` avoids is modest, and the win/loss here hinges on MariaDB 11.4's split-materialisation choice for the facet queries — that's exactly the kind of thing that flips between optimizers and data shapes. If you have an `EXPLAIN` of a feature-facet count query on trendo.cz before/after, that would settle whether this is my data being unrepresentative or a real trade-off. The AND feature itself I'd happily see merged either way.

One gap in my coverage: `isCombinationFeatureFilteringEnabled()` needs PS ≥ 9.3 plus the `combination_feature_values` flag, so the `feature_product_attribute` EXISTS branch was never exercised on my 9.1.5 shop.

kpodemski:
> [!NOTE]

Rebased on current `dev` and pushed the join based version we discussed, so the AND feature stays and the `EXISTS` rewrite goes.

> [!IMPORTANT]
> The branch was rebased and force updated:
> ```
> git fetch origin && git reset --hard origin/feature-filter-refacto
> ```

## The AND mode does not need EXISTS

Each selected value is already its own operation inside one group, and operations in a group are joined with `AND`. The adapter aliases the mapped table per operation, so `feature_product` is simply joined once per value:

```sql
LEFT JOIN ps_feature_product fp   ON (p.id_product = fp.id_product)
LEFT JOIN ps_feature_product fp_1 ON (p.id_product = fp_1.id_product)
WHERE ((fp.id_feature_value=10 AND fp_1.id_feature_value=20))
```

A product only survives if it carries both values. So `src/Adapter/MySQL.php` is back to `dev` and everything that makes the feature work, Converter, Search, Block and the templates, is untouched. I added two adapter tests pinning the AND shape and the OR shape, since nothing covered that before.

## Same behaviour

Measured on a bigger shop than last time: 100 000 products, 2 434 categories, 880 000 feature to product rows, PS 9.1.5 / PHP 8.3 / MariaDB 11.4.

Asking for two values of one feature with the filter set to `Checkbox (AND)`:

| build | products |
| --- | --- |
| SQL ground truth | 50 |
| `dev` | 438 (falls back to OR, it does not know type 4) |
| your branch | 50 |
| join version | 50 |

## And the regression is gone

| scenario | dev | your branch | join version |
| --- | ---: | ---: | ---: |
| 1 feature | 1.038 s | +4.5% | +2.8% |
| 2 features | 0.864 s | +3.4% | -6.0% |
| 3 features | 0.617 s | **+45.9%** | -4.7% |
| 1 feature + in stock | 1.440 s | **+124.9%** | -1.3% |
| category, no filter | 4.493 s | -1.7% | -2.0% |
| control, product page | 0.187 s | -5.4% | -10.0% |
| **total** | **8.640 s** | **+23.9%** | **-2.1%** |

Worse here than the 12.6% I measured on the smaller dataset, which fits the mechanism: the more facet count queries a page runs, the more often the population is re-evaluated instead of being split materialised.

## One gap I still have

`isCombinationFeatureFilteringEnabled()` needs PS 9.3 and the `combination_feature_values` flag, so the `feature_product_attribute` branch is still unexercised on my 9.1.5 shop, in both versions. If you can run the AND filter on a 9.3 shop with combination feature values on, that would close it.

Also worth saying: my dataset previously gave every product exactly one value per feature, which means an AND filter over two values of one feature returned nothing whatever the code did. A filter that always returns zero looks like it works. I fixed the dataset first, then measured, which is why the numbers above have a non zero ground truth to compare against.

kpodemski:
> [!NOTE]
> Semi-automated review. The tests below were run by Claude Code on real shops. I checked every result before posting. A human is reading and answering here, so please reply as usual.

Tested the branch on 9.1.5 and on 9.3.0, so the combination feature gap I mentioned earlier is now closed.

## The option

`Checkbox (AND)` shows on feature rows only, attribute groups keep the old list:

s1-bo-and-option

## PrestaShop 9.1.5, product level feature values

100 000 products. `F1V11` matches 238 products on its own, `F1V19` matches 208. With the filter set to `Checkbox (AND)`, asking for both gives 6, which is what the SQL says:

s2-915-and

Same URL with the filter back on plain `Checkbox`, 440 products, so the two modes really are doing different things:

s3-915-or

## PrestaShop 9.3.0, combination feature values

This is the part that was never exercised. Built with `prestashop-flashlight:nightly-nginx`, `PS_VERSION_DB` reports 9.3.0 and the `combination_feature_values` flag is on.

The data is deliberately awkward. Both products have Polyester and Wool, but:

- product 1 carries both values on the **same** combination
- product 2 carries Polyester on one combination and Wool on **another**

Polyester alone returns both, and the facet already says Wool would narrow it to 1:

s4-930-single

Polyester AND Wool returns only product 1, so the two values have to be satisfied by the same combination:

s5-930-and

## Both versions agree

| build | 9.3.0, AND on combination values |
| --- | --- |
| `dev` | 2 products, it falls back to OR |
| your original EXISTS version | 1 product |
| the join version now on the branch | 1 product |

So dropping the `EXISTS` rewrite did not change the combination behaviour, on the branch that only exists from 9.3. That was the last thing I could not check, and it holds.

kpodemski (approved):

PR #1306, opened 2026-08-28
Replace webpack/Babel/node-sass with esbuild + Dart Sass, migrate to ESLint 10

kpodemski:
| Questions         | Answers
| Description?      | Replaces the module's JavaScript build chain (webpack + Babel + node-sass) with **esbuild + Dart Sass**, and migrates linting to **ESLint 10 flat config**. Direct dev dependencies drop from 21 to 8 and the installed tree from 746 to 204 packages, which removes the bulk of the recurring Dependabot traffic on this repository. `npm ci` currently **fails on any modern Node** (`node-sass` → `node-gyp` → `ModuleNotFoundError: No module named 'distutils'`), so this also unbreaks local development. Front and back office behaviour is unchanged; bundles get smaller (`front.js` 22.2 kB → 10.7 kB, `back.js` 7.6 kB → 4.9 kB). Details below the table.
| Type?             | refacto
| BC breaks?        | no — no PHP, hook, template or public JS API change. See "Compatibility notes" below for the two developer-facing changes (Node 20.19+ required to build; bundles now target modern browsers instead of ES5).
| Deprecations?     | no
| Fixed ticket?     | None. Supersedes #1284, #1132 and #988, and makes the recurring unmergeable ESLint / Babel bumps (#1291, #1294–#1300, #1305) obsolete.
| Sponsor company   | impSolutions
| How to test?      | See "How to test" below.

---

## Why

Dependabot has been opening one PR per package on a toolchain that cannot actually move:

- **ESLint is permanently stuck at 8.** `eslint-config-prestashop@0.2.1` (last release) requires `eslint-config-airbnb-base@15`, whose peer range is `eslint ^7.32.0 || ^8.2.0`. Every ESLint 9/10 bump therefore fails `npm install` with `ERESOLVE`, and so does `@babel/eslint-parser@8` (peer `eslint ^9`). These PRs get closed and reopened forever.
- **The build is already broken locally.** `npm ci` fails on Node 20+ because `node-sass` compiles through `node-gyp`, which imports `distutils` — removed in Python 3.12.
- **CI still runs Node 14 and 16**, both long past end of life.

Bumping these one by one cannot succeed. Removing Babel, webpack and node-sass removes the problem at the source.

## What changed

**Removed** (18 dev dependencies): the 6 `@babel/*` packages, `webpack`, `webpack-cli`, `babel-loader`, `css-loader`, `style-loader`, `sass-loader`, `mini-css-extract-plugin`, `clean-webpack-plugin`, `node-sass`, `eslint-config-prestashop`, `eslint-config-airbnb-base`, `eslint-plugin-import` — plus `webpack.config.js`, `.bablerc`, `.eslintrc.js` and `.eslintignore`.

**Added** (5): `esbuild`, `sass` (Dart Sass — pure JS, no native build step), `@eslint/js`, `@stylistic/eslint-plugin`, `globals`. `chai`, `eslint` and `mocha` are kept and updated.

- `build.mjs` replaces `webpack.config.js`. It bundles the two entry points with esbuild and compiles the SCSS through a short inline Dart Sass plugin, so `views/dist/` keeps exactly the same file layout.
- `eslint.config.mjs` replaces `.eslintrc.js`: `@eslint/js` recommended plus `@stylistic` rules configured to match the style the module is already written in. The whole codebase passes with **0 errors** and no rule suppressions; three now-dead `eslint-disable` directives were removed.
- Tests run **without any transpiler**. Mocha loads the sources as native ESM, so `@babel/register` is gone. This needed three small source changes: `"type": "module"`, explicit `.js` extensions on relative imports, and `require('lodash.escaperegexp')` → `import`. All 69 assertions pass unchanged.
- `overrides` pins `diff` and `serialize-javascript` to patched versions (both reached only through Mocha's dev-time tooling, never bundled). `npm audit` reports **0 vulnerabilities**, against **22 on `dev`** (2 low, 2 moderate, 17 high, 1 critical).

## Build output

Third-party licence notices are preserved. Webpack extracted them to `views/dist/*.LICENSE.txt`; esbuild now keeps the jQuery UI Touch Punch MIT notice inline at the end of the bundle (`legalComments: 'eof'`), so the stale `.LICENSE.txt` files are deleted. The build is byte-for-byte reproducible across runs.

| | before | after |
| --- | --- | --- |
| `views/dist/front.js` | 22.2 kB | 10.7 kB |
| `views/dist/back.js` | 7.6 kB | 4.9 kB |
| build time | webpack | ~50 ms |
| `npm ci` on Node 20 | fails | ~1 s |

## Two pre-existing issues fixed along the way

1. **`views/dist` was listed in `.gitignore` while being tracked in git.** The release workflow (`PrestaShop/.github` → `build-module-artifact`) runs only `composer install` — it never runs an npm build — so the archive ships whatever `views/dist` is committed. The ignore rule was misleading; it has been removed.
2. **Nothing enforced that `views/dist` matched `_dev`.** They happen to be in sync today only because the last build and the last source change are the same August 2023 commit. A new `js-build` CI job now rebuilds and runs `git diff --exit-code -- views/dist`.

## CI and Dependabot

- `.github/workflows/js.yml`: Node matrix `20 / 22 / 24` (was `14 / 16`), `actions/checkout@v5`, `actions/setup-node@v5` with npm caching, `npm ci` instead of `npm install`, plus the new built-assets check.
- `.github/dependabot.yml`: weekly instead of daily, with **grouped updates** — npm dev minor/patch, npm dev major, npm production, composer dev, composer production — so bumps arrive as one PR per group rather than one per package. Adds the `github-actions` ecosystem, which was not being tracked at all.

## Compatibility notes

- **Building the module** now requires Node **20.19+** (ESLint 10's floor). Merchants and the release pipeline are unaffected — neither runs npm.
- **Bundles target `chrome80, firefox78, safari14, edge88`** instead of Babel's default ES5 output. The module's own code is ES6+ and there is no polyfill in the bundle either way, so this only affects browsers that could not run the source features at all. It does mean the output is no longer IE11-shaped; flagging it explicitly since the README still claims 1.7.6.0 compatibility.

## How to test

```bash
npm ci          # succeeds — on dev it fails with a node-gyp/distutils error
npm run lint    # 0 problems
npm test        # 69 passing
npm run build   # writes views/dist, ~50 ms
git diff --exit-code -- views/dist   # clean: committed assets match the sources
npm audit       # 0 vulnerabilities — on dev: 22
```

Then, in a shop with the module installed:

- **Front office** — open a category with a price/weight slider. The slider renders, the label reformats while dragging, releasing it refreshes the product list, and the loading overlay appears and disappears. Check a locale using a non-`.` decimal separator (e.g. `fr-FR`) so the CLDR formatter is exercised.
- **Back office** — Modules → Faceted search → Configure. Reorder filters by drag and drop, toggle filter checkboxes and confirm the counter updates, submit an empty template and confirm the validation alerts appear, and run both "Index all missing prices" and the attribute indexing buttons.
- Repeat the front office check with **Advanced Parameters → Performance → Smart cache for JavaScript/CSS enabled**, to confirm the assets still combine correctly.


kpodemski:
@PrestaShop/committers @tblivet need a review to simplify this part :)

Codencode:
Overall, the PR looks good to me. One thing that came up during an AI-assisted analysis of the changes is the handling of `views/dist`.

The previous webpack configuration used `CleanWebpackPlugin`, while the new esbuild script writes directly to `views/dist` without cleaning it first.

Since esbuild doesn't remove stale files from the output directory, a renamed or removed asset could remain in `views/dist` and potentially end up in a release.

Also, the new CI check:

```bash
git diff --exit-code -- views/dist
```

only detects changes to tracked files, so it wouldn't catch a newly generated untracked file.

It might therefore be safer to:

* clean `views/dist` before running the build;
* make the CI check also detect untracked files in `views/dist`, for example using `git status --porcelain -- views/dist`.

That would make the "generated assets are in sync with sources" check more robust.

Apart from this point, the changes look good to me.

kpodemski:
@Codencode I've applied the changes you suggested

Codencode (approved):

PR #1308, opened 2026-08-28
Add options to keep single-value facets visible

Touxten:
| Questions         | Answers
| Description?      | Add options to keep single-value facets visible
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | 
| How to test?      | See bellow
| Sponsor company   | @Touxten 

_How to test_

1. Go to Faceted Search → Configure.
2. Enable both new options.
3. Filter a category until only one product remains.
4. Check that:
   - Attribute groups such as Size and Color remain visible.
   - The Brand filter remains visible.
5. Disable each option and verify that its corresponding filter disappears.

### DIFF
--- a/ps_facetedsearch.php
+++ b/ps_facetedsearch.php
@@ -208,6 +208,8 @@ class Ps_Facetedsearch extends Module implements WidgetInterface
             Configuration::updateValue('PS_LAYERED_FILTER_PRICE_ROUNDING', 1);
             Configuration::updateValue('PS_LAYERED_FILTER_SHOW_OUT_OF_STOCK_LAST', 0);
             Configuration::updateValue('PS_LAYERED_FILTER_BY_DEFAULT_CATEGORY', 0);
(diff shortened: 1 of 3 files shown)


Hlavtox:
@Touxten I think this could maybe be configurable on filter level? And why only attribute, brand and not other filters?

Touxten:
@Hlavtox 

I've identified three possibilities.

**Create a single option for all types.**

To me, that isn't flexible.

**Create options for all types:** 

‘category’ 'condition' ‘extras’ 'manufacturer' ‘availability’ 'attribute_group'  ‘feature’  'weight' ‘price’

I’m worried that e-commerce merchants might get confused.

**Create options for what seems useful to the community**   


I'll see if I can add it to a specific filter

Touxten:
@Hlavtox 

Did you think about that?

Screenshot - 2026-08-28T154449 378
Screenshot - 2026-08-28T154405 309

PR #1310, opened 2026-08-29
Cache the per-shop country lookups during price indexation

boo-code:
| Questions         | Answers
| Description?      | Follow-up to @kpodemski's review note on #1261. `indexProductPrices()` runs once per product, and inside its per-shop loop it called `Country::getCountriesByIdShop()` and, a few lines below, `Country::getCountries()`. Neither is cached in core - both are plain `executeS()` calls with joins - so a full re-index repeated both once per product per shop. The first of the two became unconditional in #1261 (before that it only ran when a product had no tax rule), which is the regression that was flagged; the second was already there. Both results depend only on the shop and the language, so they are now held in statics for the life of the request, keyed on exactly that, the same way `$groups` already is a few lines above. Opened as a draft, as agreed on #1261.
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | Follow-up to PrestaShop/ps_facetedsearch#1261 (review note, no separate issue).
| How to test?      | Re-index prices (module configuration, "Index all missing prices" / a full re-index) and count the two lookups, or read the numbers below. Correctness is checked by comparing the `layered_price_index` table before and after: it must be identical, since only the number of times the country list is fetched changes. On a default 9.2 shop with 19 products and 2 shops, `getCountriesByIdShop` goes from 38 calls to 2 and `getCountries` from 38 to 1, with a byte identical index.
| Sponsor company   |

Measured on a real 9.2 install (19 products, 2 shops, 2 active countries), instrumenting both call sites and running a full re-index of every product:

```
before:  products=19  getCountriesByIdShop_calls=38  getCountries_calls=38
after:   products=19  getCountriesByIdShop_calls=2   getCountries_calls=1
```

2 and 1 are the expected cardinalities: one lookup per shop-and-language, one per language. The `layered_price_index` rows produced by the two runs are identical (38 rows, same values), so this only removes repeated queries.

The saving scales with the catalogue rather than with this shop: a re-index of N products across S shops did 2 x N x S of these queries and now does S + 1.

### DIFF
--- a/ps_facetedsearch.php
+++ b/ps_facetedsearch.php
@@ -396,6 +396,11 @@ class Ps_Facetedsearch extends Module implements WidgetInterface
     public function indexProductPrices($idProduct, $smart = true)
     {
         static $groups = null;
+        // Both country lookups below are loop invariant but sit inside the per-product path, so a
+        // full re-index repeated them once per product per shop. Neither is cached in core: both are
+        // plain executeS() calls. Held here like $groups above, for the life of the request.
+        static $shopCountries = [];
+        static $countries = [];
 
         if ($groups === null) {
             $groups = $this->getDatabase()->executeS('SELECT id_group FROM `' . _DB_PREFIX_ . 'group_reduction`');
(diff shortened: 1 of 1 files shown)


kpodemski (approved):

Hlavtox (approved):

boo-code:
Superseded by #1312, which moves price indexation into `PriceIndexer` and caches the shop countries, groups and currencies there. It deletes the body of `indexProductPrices()` that this PR optimises, so there is nothing left here to apply.

Closing rather than leaving a conflicting branch open.

boo-code:
Reopening — I closed this too early and the reasoning was wrong.

I closed it because #1312 deletes the body of `indexProductPrices()` that this patch optimises. That part is true, but it is not sufficient: #1312 bumps `config.xml` from 5.0.0 to 5.1.0, carries `upgrade/upgrade-5.1.0.php`, is milestoned 5.1.0 and currently has no reviews. `dev` and `master` are both still 5.0.0.

This PR has no milestone, so it was eligible for a 5.0.x patch. Closing it in favour of an unreviewed feature-release refactor traded something small and mergeable now for something that may take a while and may change shape — that is not mine to decide on the maintainers' behalf.

So it stands on its own until #1312 actually lands. If #1312 merges first I will close this myself as genuinely superseded, and if the maintainers would rather take the refactor as the only change here, closing it is one click and I will not reopen it again.

PR #1312, opened 2026-08-31
Refactor price indexer, add group indexation

Hlavtox:
| Questions         | Answers
| Description?      | Completely refactors the price indexation part. Fixes databases inflated to insane numbers for nothing, fixes sorting and filtering in most cases. More info below.
| Type?             | refacto
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | 
| How to test?      | 

### What this changes and improves
- Moves price indexation into a dedicated `PriceIndexer` class.
- Adds customer group support to the price index.
- **Reduces the price index size automatically without requiring configuration.** Detects whether the calculated min/max price actually depends on country or group and stores only the the required combinations:
  - 0 / 0 when neither dimension affects the price.
  - country / 0 when only country affects the price.
  - 0 / group when only group affects the price.
  - country / group when both dimensions affect the price.
- Uses IN (0, current_id) when selecting prices for the current country and group.
- Caches shop-specific countries, groups, and currencies during indexation.

### A shop where there all world countries:
- Current index size: 2 621 862 rows
- Module upgrade OK, database modified, price index still works.
- New index size: 10 638 rows

### DIFF
--- a/ps_facetedsearch.php
+++ b/ps_facetedsearch.php
@@ -365,7 +365,7 @@ class Ps_Facetedsearch extends Module implements WidgetInterface
         } elseif ($cursor == 0) {
             $this->getDatabase()->execute(
                 'DELETE psi FROM `' . _DB_PREFIX_ . 'layered_price_index` psi ' .
-                'LEFT JOIN `' . _DB_PREFIX_ . 'product_shop` ps ON (ps.`id_product` = psi.`id_product` AND ps.`active` = 1 AND ps.`visibility` IN ("both", "catalog")) ' .
+                'LEFT JOIN `' . _DB_PREFIX_ . 'product_shop` ps ON (ps.`id_product` = psi.`id_product` AND ps.`id_shop` = psi.`id_shop` AND ps.`active` = 1 AND ps.`visibility` IN (\'both\', \'catalog\')) ' .
                 'WHERE ps.`id_product` IS NULL'
             );
         }
--- a/ps_facetedsearch.php
+++ b/ps_facetedsearch.php
@@ -29,6 +29,7 @@ if (file_exists($autoloadPath)) {
 use PrestaShop\Module\FacetedSearch\Filters\Converter;
 use PrestaShop\Module\FacetedSearch\Filters\DataAccessor;
 use PrestaShop\Module\FacetedSearch\HookDispatcher;
+use PrestaShop\Module\FacetedSearch\Indexation\PriceIndexer;
 use PrestaShop\PrestaShop\Core\Module\WidgetInterface;
 
 class Ps_Facetedsearch extends Module implements WidgetInterface
@@ -51,22 +52,6 @@ class Ps_Facetedsearch extends Module implements WidgetInterface
      * @var int
      */
     const LOCK_TEMPLATE_CREATION = 20000;
-
-    /**
-     * US iso code, used to prevent taxes usage while computing prices
-     *
-     * @var array
-     */
-    const ISO_CODE_TAX_FREE = [
-        'US',
-    ];
-
-    /**
-     * Number of digits for MySQL DECIMAL
-     *
-     * @var int
-     */
-    const DECIMAL_DIGITS = 6;
 
     /**
      * @var array List of controllers supported by this module
@@ -93,11 +78,16 @@ class Ps_Facetedsearch extends Module implements WidgetInterface
      */
     private $hookDispatcher;
 
+    /**
+     * @var PriceIndexer
+     */
+    private $priceIndexer;
+
     public function __construct()
     {
         $this->name = 'ps_facetedsearch';
         $this->tab = 'front_office_features';
-        $this->version = '5.0.0';
+        $this->version = '5.1.0';
         $this->author = 'PrestaShop';
         $this->need_instance = 0;
         $this->bootstrap = true;
@@ -369,8 +359,15 @@ class Ps_Facetedsearch extends Module implements WidgetInterface
      */
     public function fullPricesIndexProcess($cursor = 0, $ajax = false, $smart = false)
     {
+        // Prepare the price index before the first batch while preserving valid rows during smart indexing.
         if ($cursor == 0 && !$smart) {
             $this->rebuildPriceIndexTable();
+        } elseif ($cursor == 0) {
+            $this->getDatabase()->execute(
+                'DELETE psi FROM `' . _DB_PREFIX_ . 'layered_price_index` psi ' .
+                'LEFT JOIN `' . _DB_PREFIX_ . 'product_shop` ps ON (ps.`id_product` = psi.`id_product` AND ps.`active` = 1 AND ps.`visibility` IN ("both", "catalog")) ' .
+                'WHERE ps.`id_product` IS NULL'
+            );
         }
 
         return $this->indexPrices($cursor, true, $ajax, $smart);
@@ -395,211 +392,12 @@ class Ps_Facetedsearch extends Module implements WidgetInterface
      */
     public function indexProductPrices($idProduct, $smart = true)
     {
-        static $groups = null;
-
-        if ($groups === null) {
-            $groups = $this->getDatabase()->executeS('SELECT id_group FROM `' . _DB_PREFIX_ . 'group_reduction`');
-            if (!$groups) {
-                $groups = [];
-            }
+        // Reuse one indexer instance so its shop dimension caches survive the complete indexing process.
+        if ($this->priceIndexer === null) {
+            $this->priceIndexer = new PriceIndexer();
         }
 
-        $shopList = Shop::getShops(false, null, true);
-
-        foreach ($shopList as $idShop) {
-            $currencyList = Currency::getCurrencies(false, true, true);
-
-            $minPrice = [];
-            $maxPrice = [];
-
-            if ($smart) {
-                $this->getDatabase()->execute('DELETE FROM `' . _DB_PREFIX_ . 'layered_price_index` WHERE `id_product` = ' . (int) $idProduct . ' AND `id_shop` = ' . (int) $idShop);
-            }
-
-            $taxRatesByCountry = $this->getDatabase()->executeS(
-                'SELECT t.rate rate, tr.id_country, c.iso_code ' .
-                'FROM `' . _DB_PREFIX_ . 'product_shop` p ' .
-                'LEFT JOIN `' . _DB_PREFIX_ . 'tax_rules_group` trg ON  ' .
-                '(trg.id_tax_rules_group = p.id_tax_rules_group AND p.id_shop = ' . (int) $idShop . ') ' .
-                'LEFT JOIN `' . _DB_PREFIX_ . 'tax_rule` tr ON (tr.id_tax_rules_group = trg.id_tax_rules_group) ' .
-                'LEFT JOIN `' . _DB_PREFIX_ . 'tax` t ON (t.id_tax = tr.id_tax AND t.active = 1) ' .
-                'JOIN `' . _DB_PREFIX_ . 'country` c ON (tr.id_country=c.id_country AND c.active = 1) ' .
-                'WHERE id_product = ' . (int) $idProduct . ' ' .
-                'GROUP BY id_product, tr.id_country'
-            );
-
-            // When tax is not applied to indexed prices, drop the per-country tax rates entirely.
-            if (!Configuration::get('PS_LAYERED_FILTER_PRICE_USETAX')) {
-                $taxRatesByCountry = [];
-            }
-
-            // Always index every active country of the shop. Countries that have no specific tax
-            // rule for this product (or when the "use tax" option is off) are indexed with a 0% rate.
-            // Without this, the price sort - an INNER JOIN on layered_price_index.id_country - returns
-            // no products at all for customers whose country was not indexed. This makes the
-            // with-tax-rules path consistent with the path already used for products without any tax
-            // rule. See https://github.com/PrestaShop/ps_facetedsearch/issues/1206
-            $indexedCountryIds = array_column($taxRatesByCountry, 'id_country');
-            $shopCountries = Country::getCountriesByIdShop($idShop, $this->getContext()->language->id);
-            foreach ($shopCountries as $country) {
-                if (!empty($country['active']) && !in_array($country['id_country'], $indexedCountryIds)) {
-                    $taxRatesByCountry[] = [
-                        'rate' => 0,
-                        'id_country' => $country['id_country'],
-                        'iso_code' => $country['iso_code'],
-                    ];
-                }
-            }
-
-            $productMinPrices = $this->getDatabase()->executeS(
-                'SELECT id_shop, id_currency, id_country, id_group, from_quantity
-                FROM `' . _DB_PREFIX_ . 'specific_price`
-                WHERE id_product = ' . (int) $idProduct . ' AND id_shop IN (0,' . (int) $idShop . ')'
-            );
-
-            $countries = Country::getCountries($this->getContext()->language->id, true, false, false);
-            foreach ($countries as $country) {
-                $idCountry = $country['id_country'];
-
-                // Get price by currency & country, without reduction!
-                foreach ($currencyList as $currency) {
-                    $price = Product::priceCalculation(
-                        $idShop,
-                        (int) $idProduct,
-                        null,
-                        $idCountry,
-                        0,
-                        '',
-                        $currency['id_currency'],
-                        0,
-                        0,
-                        false,
-                        6, // Decimals
-                        false,
-                        false,
-                        true,
-                        $specificPriceOutput,
-                        true
-                    );
-
-                    $minPrice[$idCountry][$currency['id_currency']] = $price;
-                    $maxPrice[$idCountry][$currency['id_currency']] = $price;
-                }
-
-                foreach ($productMinPrices as $specificPrice) {
-                    foreach ($currencyList as $currency) {
-                        if ($specificPrice['id_currency'] &&
-                            $specificPrice['id_currency'] != $currency['id_currency']
-                        ) {
-                            continue;
-                        }
-
-                        $price = Product::priceCalculation(
-                            $idShop,
-                            (int) $idProduct,
-                            null,
-                            $idCountry,
-                            0,
-                            '',
-                            $currency['id_currency'],
-                            (int) $specificPrice['id_group'],
-                            $specificPrice['from_quantity'],
-                            false,
-                            6,
-                            false,
-                            true,
-                            true,
-                            $specificPriceOutput,
-                            true
-                        );
-
-                        if ($price > $maxPrice[$idCountry][$currency['id_currency']]) {
-                            $maxPrice[$idCountry][$currency['id_currency']] = $price;
-                        }
-
-                        if ($price == 0) {
-                            continue;
-                        }
-
-                        if (null === $minPrice[$idCountry][$currency['id_currency']] || $price < $minPrice[$idCountry][$currency['id_currency']]) {
-                            $minPrice[$idCountry][$currency['id_currency']] = $price;
-                        }
-                    }
-                }
-
-                foreach ($groups as $group) {
-                    foreach ($currencyList as $currency) {
-                        $price = Product::priceCalculation(
-                            $idShop,
-                            (int) $idProduct,
-                            null,
-                            (int) $idCountry,
-                            0,
-                            '',
-                            (int) $currency['id_currency'],
-                            (int) $group['id_group'],
-                            0,
-                            false,
-                            6,
-                            false,
-                            true,
-                            true,
-                            $specificPriceOutput,
-                            true
-                        );
-
-                        if (!isset($maxPrice[$idCountry][$currency['id_currency']])) {
-                            $maxPrice[$idCountry][$currency['id_currency']] = 0;
-                        }
-
-                        if (!isset($minPrice[$idCountry][$currency['id_currency']])) {
-                            $minPrice[$idCountry][$currency['id_currency']] = null;
-                        }
-
-                        if ($price == 0) {
-                            continue;
-                        }
-
-                        if (null === $minPrice[$idCountry][$currency['id_currency']] || $price < $minPrice[$idCountry][$currency['id_currency']]) {
-                            $minPrice[$idCountry][$currency['id_currency']] = $price;
-                        }
-
-                        if ($price > $maxPrice[$idCountry][$currency['id_currency']]) {
-                            $maxPrice[$idCountry][$currency['id_currency']] = $price;
-                        }
-                    }
-                }
-            }
-
-            $values = [];
-            foreach ($taxRatesByCountry as $taxRateByCountry) {
-                $taxRate = $taxRateByCountry['rate'];
-                $idCountry = $taxRateByCountry['id_country'];
-                foreach ($currencyList as $currency) {
-                    $minPriceValue = array_key_exists($idCountry, $minPrice) ? $minPrice[$idCountry][$currency['id_currency']] : 0;
-                    $maxPriceValue = array_key_exists($idCountry, $maxPrice) ? $maxPrice[$idCountry][$currency['id_currency']] : 0;
-                    if (!in_array($taxRateByCountry['iso_code'], self::ISO_CODE_TAX_FREE)) {
-                        $minPriceValue = Tools::ps_round($minPriceValue * (100 + $taxRate) / 100, self::DECIMAL_DIGITS);
-                        $maxPriceValue = Tools::ps_round($maxPriceValue * (100 + $taxRate) / 100, self::DECIMAL_DIGITS);
-                    }
-
-                    $values[] = '(' . (int) $idProduct . ',
-                        ' . (int) $currency['id_currency'] . ',
-                        ' . $idShop . ',
-                        ' . (float) $minPriceValue . ',
-                        ' . (float) $maxPriceValue . ',
-                        ' . (int) $idCountry . ')';
-                }
-            }
-
-            if (!empty($values)) {
-                $this->getDatabase()->execute(
-                    'INSERT INTO `' . _DB_PREFIX_ . 'layered_price_index` (id_product, id_currency, id_shop, price_min, price_max, id_country)
-                     VALUES ' . implode(',', $values) . '
-                     ON DUPLICATE KEY UPDATE id_product = id_product' // Avoid duplicate keys
-                );
-            }
-        }
+        $this->priceIndexer->indexProductPrices($idProduct, $smart);
     }
 
     /**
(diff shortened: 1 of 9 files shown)


Codencode:
I really like the direction of this refactor. Computing the full country/group price matrix first and then compacting it based on the actual resulting prices seems more robust than trying to infer dependencies only from `specific_price`, and it also naturally covers group reductions and tax differences.

I noticed two points that might be worth checking before merging:

1. **US tax handling**

The previous indexer explicitly excluded the US from tax application through `ISO_CODE_TAX_FREE = ['US']`.

With the new implementation, `getTaxRatesByCountry()` groups tax rules by country and the resulting rate is applied during price computation. Since US taxes may depend on state/ZIP code, I'm wondering whether removing the previous special handling is intentional.

Could this potentially result in a state-specific tax rate being used as a generic US tax rate during indexation?

2. **Current customer group in multistore**

The reader currently uses:

`$this->getContext()->customer->id_default_group`

to filter the index with:

`psi.id_group IN (0, current_group)`

However, the core can resolve a different effective group for the current shop when the customer's default group is not associated with that shop (`Group::getCurrent()` handles this fallback).

Since the new index only contains groups associated with the shop, could this create a case where no indexed row matches for that customer in multistore?

Maybe using the effective current group resolved by the core would be safer here.

Apart from these two points, the overall approach looks very promising to me.

Hlavtox:
@Codencode Thanks for pointing these out! I will answer in points.

1) I removed the manual tax-rule lookup entirely. The indexer now passes the configured `PS_LAYERED_FILTER_PRICE_USETAX` value directly to `Product::priceCalculation()`, so tax handling is delegated to the core pricing logic. There is no longer a hardcoded US exception, nor a risk of selecting an arbitrary state-specific rate through a country-level GROUP BY. It completely respects the core calculation.

2) I also changed the reader to use Group::getCurrent()->id, including the specific-price join, so the effective shop-associated customer group is used consistently in multistore.

boo-code:
Looked at it, as asked on PrestaShop/PrestaShop#16735. Two things.

### The group axis is there, but the tax basis is still shop-wide

`getPriceMatrix()` iterates groups and hands the same `$useTax` to every one of them:

```php
$useTax = (bool) Configuration::get('PS_LAYERED_FILTER_PRICE_USETAX', null, null, $idShop);
...
foreach ($groups as $group) {
    $idGroup = (int) $group['id_group'];
    $prices[$idCountry][$idGroup] = $this->calculatePriceRange($idProduct, $idShop, $idCurrency, $idCountry, $idGroup, $specificPriceQuantities, $useTax);
}
```

But whether a customer sees prices with or without tax is itself a per-group property — `ps_group.price_display_method`, reachable as `Group::getPriceDisplayMethod($idGroup)` (`PS_TAX_EXC = 1`, `PS_TAX_INC = 0`). So after this refactor the index can express a per-group price, and still stores every group's range in one tax basis chosen by a module setting.

That is the whole of #16735: on the shop I measured there, a group displaying tax-excluded prices sees `€19.12 … €35.90` on the page while the slider is built from the tax-included index and offers `22 – 44` — a minimum that already excludes the cheapest visible product.

Deriving it inside the loop rather than above it would close that:

```php
$useTax = Group::getPriceDisplayMethod($idGroup) == PS_TAX_EXC ? false : true;
```

It also composes with the dependency detection rather than fighting it: on a shop whose groups all display the same way the computed ranges stay identical and the group dimension still collapses to `0`, so the index-size win is untouched. It only expands where the groups genuinely differ, which is exactly where the current index is wrong.

If you would rather keep `PS_LAYERED_FILTER_PRICE_USETAX` as an override, the same line works as a default when the setting is unset.

### #1310 is superseded by this

My open #1310 caches the two `Country::` lookups inside `indexProductPrices()`. This PR deletes that method's body wholesale and reimplements the caching in `PriceIndexer`, so there is nothing left for it to apply to. I will close it rather than leave you a conflicting branch to reason about.

Separately, #1260 touches `ps_facetedsearch.php`, `src/Adapter/MySQL.php` and `tests/php/FacetedSearch/Adapter/MySQLTest.php`, so it will need a rebase once this lands. That one is mine to deal with.

boo-code:
Correcting the last paragraph of my comment above: I said I would close #1310 as superseded, did, and have now reopened it.

The supersession is real in code — this PR deletes the method that one optimises — but it only takes effect when this lands. This is milestoned 5.1.0 and bumps `config.xml` to 5.1.0 with an upgrade script, while `dev` and `master` are both 5.0.0; #1310 has no milestone and so is eligible for a 5.0.x patch. Closing it early would have removed that option on the assumption that this merges, which is not my call to make.

I will close it myself once this is merged. Nothing about the review above changes.

Hlavtox:
Ping @kpodemski

Codencode (dismissed):

mattgoud (commented):
Tested locally on a 9.2.x shop with the branch merged into `dev`. The refactor does what it says and it fixes a real bug, so I am in favour of it landing in 5.1.0. Three things are worth settling first, one of which makes the migration a no-op in practice.

## What I verified

| Check | Result |
|---|---|
| Upgrade 5.0.0 to 5.1.0 | Column added, PK extended, existing rows preserved as `id_group = 0`, front office keeps working immediately because the join is `id_group IN (0, x)`. No downtime. |
| Fresh install | `rebuildPriceIndexTable()` is the only `CREATE TABLE` site, so a new install gets `id_group`. Verified through a module reset. |
| No regression | Front office output is byte identical between `dev` and this branch (sliders, price sorts, filter counts, product ids) on the default shop with 241 active countries and `PS_LAYERED_FILTER_PRICE_USETAX = 1`. |
| The actual fix | With a 20% reduction on the Customer group: `dev` offers that customer a 10 to 44 slider, this branch offers 8 to 35, which is their real range. `dev` never indexed `ps_group.reduction` at all, it only looped over groups present in `ps_group_reduction`. |
| Cache key | Necessary and working: the block hash now differs per group (`3059d2bb...` vs `54cdab5d...`). Without it the group aware index would have been defeated by `ps_layered_filter_block`. |
| Compaction | With no country dependency: 4579 rows down to 19, and the front office still resolves for every country through `id_country = 0`. |

## 1. `PS_LAYERED_INDEXED = 0` does not mark the index as stale

The upgrade script comment says it marks the preserved index as stale so a normal full rebuild can compact it, but nothing reads the flag for that. Its only consumer is `_dev/back/index.js`:

```js
if (typeof PS_LAYERED_INDEXED !== 'undefined' && PS_LAYERED_INDEXED) {
  $('#url-indexe').click();
  $('#full-index').click();
}
```

Neither `#url-indexe` nor `#full-index` exists in `views/templates/admin/manage.tpl`, so both selectors are dead and the flag drives nothing. There is also no template branch showing an "index out of date" warning tied to it.

The consequence is that an upgraded shop keeps its pre-5.1.0 rows indefinitely. Because the new join is tolerant, every filter and sort keeps working silently, so no merchant has a reason to go and rebuild. Both headline benefits of the PR, group correct prices and the smaller index, only land on shops that happen to open Design > Faceted search and click the indexer themselves.

Either rebuild from the upgrade script (batched, the way `fullPricesIndexProcess` already batches), or add a real warning that the price index needs a rebuild. As it stands the migration is a schema change with no data migration.

## 2. Indexation cost now scales with the number of customer groups, and the size win is conditional

`getGroupsByShop()` returns every group associated with the shop, and `getPriceMatrix()` computes `countries x groups x currencies` unconditionally. `dev` was flat in the number of groups because it only looped groups with a `ps_group_reduction` row.

Measured on 19 products, 241 active countries, 1 currency, `USETAX = 1`:

| Groups | `dev` | This branch | Rows (both) |
|---|---|---|---|
| 3 | 0.44 s | 0.84 s | 4579 |
| 12 | 0.41 s | 3.0 s | 4579 |

No rows saved, roughly 7x the indexation time at 12 groups. The reason is that a single tax rule makes the calculated price country dependent, so `dependsOnCountry` is true and the country axis does not collapse at all. The 2 621 862 to 10 638 figure in the description comes from a shop where no country affects the price, which I reproduced by setting `PS_LAYERED_FILTER_PRICE_USETAX = 0`: 4579 rows down to 19. That is a genuine win, it is just narrower than the description suggests, and B2B shops with many groups pay the cost without getting it.

Two mitigations that keep the fix intact:

* Skip the group loop entirely when nothing can make the price group dependent (no `ps_group.reduction`, no `ps_group_reduction` row, no `specific_price` with `id_group != 0`, uniform `price_display_method`) and compute once with the default group. Dependency detection then collapses the axis exactly as it does today, so the output is unchanged.
* Compute one representative country per distinct tax rate rather than all 241, then fan the result out to the countries sharing that rate. Typically 1 to 5 calls instead of 241, and the stored rows are identical.

## 3. Dropping the US tax exemption is an undocumented behaviour change

`ISO_CODE_TAX_FREE` and `DECIMAL_DIGITS` are removed. They are public constants on the module class, so strictly speaking that is a BC break on a PR marked `BC breaks: no`. More importantly the exemption itself is gone, and that is observable. With a 20% US tax rule added to the product's tax rules group:

* `dev` indexes US at `19.120000`, tax skipped through `ISO_CODE_TAX_FREE`
* this branch indexes US at `22.944000`, tax applied like any other country

I think applying the tax is the right call, a merchant who enables "use tax" for the index presumably wants it applied. But it should be stated in the description rather than discovered by a US shop after upgrading, and the two constants are cheap to keep as deprecated.

## Nits

* No unit test for `PriceIndexer`, and 365 new lines is where they would pay off most. `compactPriceMatrix()` is pure and takes four cases (country only, group only, both, neither). The PR only updates existing expectations.
* Schema drift between the two paths, verified on both: the upgrade adds `id_group INT NOT NULL DEFAULT 0`, `rebuildPriceIndexTable()` declares `id_group INT NOT NULL` with no default.
* The new smart mode DELETE uses `ps.visibility IN ("both", "catalog")`. Single quotes elsewhere in the module and safe under `ANSI_QUOTES`.
* That same DELETE does not constrain `ps.id_shop = psi.id_shop`, so in multistore a product disabled in one shop keeps its rows for that shop. Still an improvement over no cleanup at all.
* `calculatePriceRange()`: `$minPrice === null` can never be true, `$minPrice` is seeded from `priceCalculation()`. Dead branch.
* The `config.xml` and `$this->version` bump to 5.1.0 is redundant now that #1325 has landed on `dev`. It merges cleanly, it is just noise to drop on the next rebase.

## On the `price_display_method` point raised above

I reproduced it. A group set to display tax excluded sees prices from 9.00 to 35.90 while the slider is built from the tax included index and offers 10 to 44. It is pre-existing rather than a regression from this PR, but this PR is the natural place to fix it since it is the one introducing the group axis, and per the dependency detection it costs nothing on shops whose groups all display the same way.

mattgoud (commented):
Follow up on my review above with the concrete changes. Four of the points are small enough to be committable suggestions, the other two are comments where a suggestion would not fit.

The two structural points from my review are deliberately not here: the `PS_LAYERED_INDEXED` flag being a no-op needs a design decision (batched rebuild from the upgrade script, or a real "index out of date" warning, and the warning would touch `manage.tpl` and `_dev/back/index.js` which are not in this diff), and the country axis needs restructuring rather than a tweak.

On that second one, one more measurement worth having: in the scenario with a reduction on one group, this branch produces **54 948 rows** where `dev` produces 4 579, for 19 products. Extrapolated to a 10k product catalogue that is around 29M rows. So on the shops this PR is meant to fix, the index grows substantially rather than shrinking. I also checked whether it could be avoided by storing only the deviating groups plus an `id_group = 0` fallback row: it cannot, because `id_group IN (0, x)` would then match two rows and the INNER JOIN would duplicate products. The all or nothing behaviour is forced by the current join, which is why I am flagging it as a design question and not a nit.

src/Indexation/PriceIndexer.php
```diff
@@ -0,0 +1,365 @@
...
+            self::$groupsByShop[$idShop] = Db::getInstance()->executeS(
+                'SELECT g.id_group
+                FROM `' . _DB_PREFIX_ . 'group` g
+                INNER JOIN `' . _DB_PREFIX_ . 'group_shop` gs ON (gs.id_group = g.id_group)
+                WHERE gs.id_shop = ' . (int) $idShop
+            ) ?: [];
+        }
+
+        return self::$groupsByShop[$idShop];
+    }
```
mattgoud, line 168: This is the mitigation for the group scaling cost I measured in my review. Loading every group means `getPriceMatrix()` computes `countries x groups x currencies` even when no group can possibly change the price, which is the case on the large majority of shops.

Gating it keeps the fix intact: when nothing in the shop can make a price group dependent, one representative group yields the same range for all of them, and the dependency detection already stores `id_group = 0` in that situation, so the resulting index is unchanged.

I implemented and benchmarked exactly this, comparing an MD5 fingerprint of the whole `layered_price_index` table before and after, on 19 products / 241 active countries / 12 groups / 1 currency:

| Scenario | PR as is | With this change | Index |
|---|---|---|---|
| No group specific pricing | 2.65 s | **0.44 s** | identical |
| `ps_group.reduction` set | 3.02 s | 3.23 s | identical |
| `ps_group_reduction` row | 2.84 s | 2.71 s | identical |
| `specific_price` with `id_group != 0` | 2.83 s | 2.70 s | identical |
| `USETAX = 0` (full compaction) | 2.59 s | **0.32 s** | identical |

Byte identical index in all five, roughly 6x faster on the common case, and no measurable cost when the groups genuinely matter. It also brings indexation time back in line with `dev`, which was flat in the number of groups.

The gate is deliberately conservative, any one of the three signals keeps the full group matrix. It runs once per shop because it sits inside the existing cache miss, so no extra caching is needed.

```suggestion
    private function getGroupsByShop($idShop)
    {
        if (!isset(self::$groupsByShop[$idShop])) {
            // Load every group explicitly associated with the indexed shop because groups have no active flag.
            $groups = Db::getInstance()->executeS(
                'SELECT g.id_group
                FROM `' . _DB_PREFIX_ . 'group` g
                INNER JOIN `' . _DB_PREFIX_ . 'group_shop` gs ON (gs.id_group = g.id_group)
                WHERE gs.id_shop = ' . (int) $idShop
            ) ?: [];

            // When nothing in this shop can make a price depend on the customer group, every group yields
            // the same range, so a single representative group is enough. This avoids multiplying every
            // price calculation by the number of groups; the dependency detection below stores
            // id_group = 0 in that case anyway, so the resulting index is unchanged.
            if (count($groups) > 1 && !$this->hasGroupSpecificPricing($idShop)) {
                $groups = [reset($groups)];
            }

            self::$groupsByShop[$idShop] = $groups;
        }

        return self::$groupsByShop[$idShop];
    }

    /**
     * Tell whether anything in a shop can make a calculated price depend on the customer group.
     *
     * @param int $idShop
     *
     * @return bool
     */
    private function hasGroupSpecificPricing($idShop)
    {
        $hasGroupReduction = (bool) Db::getInstance()->getValue(
            'SELECT 1
            FROM `' . _DB_PREFIX_ . 'group` g
            INNER JOIN `' . _DB_PREFIX_ . 'group_shop` gs ON (gs.id_group = g.id_group)
            WHERE gs.id_shop = ' . (int) $idShop . ' AND g.reduction != 0'
        );

        $hasCategoryGroupReduction = (bool) Db::getInstance()->getValue(
            'SELECT 1 FROM `' . _DB_PREFIX_ . 'group_reduction`'
        );

        $hasGroupSpecificPrice = (bool) Db::getInstance()->getValue(
            'SELECT 1
            FROM `' . _DB_PREFIX_ . 'specific_price`
            WHERE id_group != 0 AND id_shop IN (0, ' . (int) $idShop . ')'
        );

        return $hasGroupReduction || $hasCategoryGroupReduction || $hasGroupSpecificPrice;
    }
```

ps_facetedsearch.php
```diff
@@ -369,8 +359,15 @@ public function indexAttributeGroup()
      */
     public function fullPricesIndexProcess($cursor = 0, $ajax = false, $smart = false)
     {
+        // Prepare the price index before the first batch while preserving valid rows during smart indexing.
         if ($cursor == 0 && !$smart) {
             $this->rebuildPriceIndexTable();
+        } elseif ($cursor == 0) {
+            $this->getDatabase()->execute(
+                'DELETE psi FROM `' . _DB_PREFIX_ . 'layered_price_index` psi ' .
+                'LEFT JOIN `' . _DB_PREFIX_ . 'product_shop` ps ON (ps.`id_product` = psi.`id_product` AND ps.`id_shop` = psi.`id_shop` AND ps.`active` = 1 AND ps.`visibility` IN ("both", "catalog")) ' .
```
mattgoud: Two small things on this join.

`"both"` and `"catalog"` are double quoted, which is the only place in the module that does so and breaks under the `ANSI_QUOTES` sql_mode, where double quotes are identifiers rather than string literals.

The join also does not constrain the shop, so a row is kept as long as the product is active and visible in *any* shop. In multistore, a product disabled in shop 1 but active in shop 2 keeps its stale shop 1 rows. Adding `ps.id_shop = psi.id_shop` makes the cleanup per shop, which is what the rest of the indexer does.

```suggestion
                'LEFT JOIN `' . _DB_PREFIX_ . 'product_shop` ps ON (ps.`id_product` = psi.`id_product` AND ps.`id_shop` = psi.`id_shop` AND ps.`active` = 1 AND ps.`visibility` IN (\'both\', \'catalog\')) ' .
```

ps_facetedsearch.php
```diff
@@ -1454,7 +1252,8 @@ public function rebuildPriceIndexTable()
             `price_min` DECIMAL(20, 6) NOT NULL,
             `price_max` DECIMAL(20, 6) NOT NULL,
             `id_country` INT NOT NULL,
-            PRIMARY KEY (`id_product`, `id_currency`, `id_shop`, `id_country`),
+            `id_group` INT NOT NULL,
```
mattgoud, line 1255: `upgrade-5.1.0.php` adds the column as `id_group INT NOT NULL DEFAULT 0` while this declares it without a default, so an upgraded shop and a fresh install end up with different column definitions. I verified both paths on a local shop, `DESC ps_layered_price_index` reports `Default: 0` after the upgrade and `Default: NULL` after a module reset.

Harmless today because every INSERT lists `id_group` explicitly, but it is free to align, and the default is the useful one for anything writing to the table without the column.

```suggestion
            `id_group` INT NOT NULL DEFAULT 0,
```

src/Indexation/PriceIndexer.php
```diff
@@ -0,0 +1,365 @@
...
+
+            if ($price > $maxPrice) {
+                $maxPrice = $price;
+            }
+
+            if ($price == 0) {
+                continue;
+            }
+
+            if ($minPrice === null || $price < $minPrice) {
```
mattgoud, line 310: `$minPrice` is seeded from the `priceCalculation()` call above and is never reassigned to null, so `$minPrice === null` can never be true. It is a leftover from the old code, where `$minPrice` really was initialised to null.

```suggestion
            if ($price < $minPrice) {
```

ps_facetedsearch.php
```diff
@@ -52,22 +53,6 @@ class Ps_Facetedsearch extends Module implements WidgetInterface
      */
     const LOCK_TEMPLATE_CREATION = 20000;
 
-    /**
-     * US iso code, used to prevent taxes usage while computing prices
-     *
-     * @var array
-     */
-    const ISO_CODE_TAX_FREE = [
```
mattgoud, line 60: These are public constants on the module class, so removing them is technically a BC break on a PR marked `BC breaks: no`. Nothing in core or in the native modules references them, I checked, but third party code can. Keeping them with an `@deprecated` tag and scheduling removal for the next major would cost nothing.

The behaviour change matters more than the constants themselves and is worth a line in the description. `ISO_CODE_TAX_FREE` exempted the US from the indexed tax, and that exemption is gone now that the tax basis is delegated to `Product::priceCalculation()`. Reproduced locally by adding a 20% US tax rule to the product's tax rules group:

* `dev` indexes US at `19.120000`, tax skipped
* this branch indexes US at `22.944000`, tax applied like any other country

I think applying it is the right call, a merchant enabling "use tax" for the index presumably wants it applied. But a US shop should not have to discover it after upgrading.

src/Indexation/PriceIndexer.php
```diff
@@ -0,0 +1,365 @@
...
+ */
+
+namespace PrestaShop\Module\FacetedSearch\Indexation;
+
+use Configuration;
+use Db;
+use Product;
+use Shop;
+
+class PriceIndexer
```
mattgoud, line 28: 365 new lines with no unit test, while the PR only updates existing expectations in `MySQLTest` and `BlockTest`.

`compactPriceMatrix()` is the natural candidate: it is pure, it takes a plain array in and returns a plain array out, and it has exactly four cases worth pinning down (depends on country only, on group only, on both, on neither). That is the piece of logic the whole index size claim rests on, and a regression there would silently produce a wrong index rather than an error.

`calculatePriceRange()` is harder to unit test since it calls into `Product::priceCalculation()`, so I would not ask for that one.

src/Indexation/PriceIndexer.php
mattgoud: Correcting my own suggestion: it is incomplete, and in a way that matters if the `price_display_method` point raised in this thread gets fixed.

My gate tests three signals (`ps_group.reduction`, `ps_group_reduction`, group scoped `specific_price`). Those are the only things that can make a price group dependent **in the code as it stands**, which is why the index came out byte identical in the five scenarios I measured. But `price_display_method` is a fourth one, and it is currently invisible to `calculatePriceRange()` only because the tax basis is read once per shop.

So if the per group tax basis lands, my gate silently re-breaks it. Measured on a shop with group 3 set to display tax excluded, 2 countries, product 1:

| Variant | Rows | Group 3 range for FR |
|---|---|---|
| Per group tax basis alone | 114 | `19.12` to `23.90`, correct |
| Per group tax basis + my gate as posted | 38 | collapsed to `id_group = 0`, `22.94` to `28.68`, wrong |
| Per group tax basis + the gate below | 114 | `19.12` to `23.90`, correct |

Adding a uniform `price_display_method` check fixes it and costs nothing on shops where every group displays the same way, which is the default. Re-measured on 19 products / 241 countries / 12 groups: still byte identical to this branch in all five scenarios, and 2.67 s down to 0.51 s on the common case.

Worth taking whichever way the display method discussion goes, since the check is cheap and makes the gate correct by construction rather than correct by coincidence.

```suggestion
    private function getGroupsByShop($idShop)
    {
        if (!isset(self::$groupsByShop[$idShop])) {
            // Load every group explicitly associated with the indexed shop because groups have no active flag.
            $groups = Db::getInstance()->executeS(
                'SELECT g.id_group
                FROM `' . _DB_PREFIX_ . 'group` g
                INNER JOIN `' . _DB_PREFIX_ . 'group_shop` gs ON (gs.id_group = g.id_group)
                WHERE gs.id_shop = ' . (int) $idShop
            ) ?: [];

            // When nothing in this shop can make a price depend on the customer group, every group yields
            // the same range, so a single representative group is enough. This avoids multiplying every
            // price calculation by the number of groups; the dependency detection below stores
            // id_group = 0 in that case anyway, so the resulting index is unchanged.
            if (count($groups) > 1 && !$this->hasGroupSpecificPricing($idShop)) {
                $groups = [reset($groups)];
            }

            self::$groupsByShop[$idShop] = $groups;
        }

        return self::$groupsByShop[$idShop];
    }

    /**
     * Tell whether anything in a shop can make a calculated price depend on the customer group.
     *
     * @param int $idShop
     *
     * @return bool
     */
    private function hasGroupSpecificPricing($idShop)
    {
        $hasGroupReduction = (bool) Db::getInstance()->getValue(
            'SELECT 1
            FROM `' . _DB_PREFIX_ . 'group` g
            INNER JOIN `' . _DB_PREFIX_ . 'group_shop` gs ON (gs.id_group = g.id_group)
            WHERE gs.id_shop = ' . (int) $idShop . ' AND g.reduction != 0'
        );

        $hasCategoryGroupReduction = (bool) Db::getInstance()->getValue(
            'SELECT 1 FROM `' . _DB_PREFIX_ . 'group_reduction`'
        );

        $hasMixedPriceDisplayMethod = (bool) Db::getInstance()->getValue(
            'SELECT 1
            FROM `' . _DB_PREFIX_ . 'group` g
            INNER JOIN `' . _DB_PREFIX_ . 'group_shop` gs ON (gs.id_group = g.id_group)
            WHERE gs.id_shop = ' . (int) $idShop . '
            GROUP BY gs.id_shop
            HAVING COUNT(DISTINCT g.price_display_method) > 1'
        );

        $hasGroupSpecificPrice = (bool) Db::getInstance()->getValue(
            'SELECT 1
            FROM `' . _DB_PREFIX_ . 'specific_price`
            WHERE id_group != 0 AND id_shop IN (0, ' . (int) $idShop . ')'
        );

        return $hasGroupReduction
            || $hasCategoryGroupReduction
            || $hasGroupSpecificPrice
            || $hasMixedPriceDisplayMethod;
    }

    /**
     * Return active, non-deleted currencies associated with a shop.
     *
     * @param int $idShop
     *
     * @return array
     */
    private function getCurrenciesByShop($idShop)
    {
        if (!isset(self::$currenciesByShop[$idShop])) {
            // Load only active, non-deleted currencies explicitly associated with the indexed shop.
            self::$currenciesByShop[$idShop] = Db::getInstance()->executeS(
                'SELECT c.id_currency
                FROM `' . _DB_PREFIX_ . 'currency` c
                INNER JOIN `' . _DB_PREFIX_ . 'currency_shop` cs ON (cs.id_currency = c.id_currency)
                WHERE c.active = 1 AND c.deleted = 0 AND cs.id_shop = ' . (int) $idShop
            ) ?: [];
        }

        return self::$currenciesByShop[$idShop];
    }

    /**
     * Return all specific-price quantity thresholds that can affect a product in a shop.
     *
     * @param int $idProduct
     * @param int $idShop
     *
     * @return array
     */
    private function getSpecificPriceQuantities($idProduct, $idShop)
    {
        $specificPrices = Db::getInstance()->executeS(
            'SELECT DISTINCT from_quantity
            FROM `' . _DB_PREFIX_ . 'specific_price`
            WHERE id_product = ' . (int) $idProduct . ' AND id_shop IN (0,' . (int) $idShop . ')'
        );

        // Return scalar quantities so the calculation loop remains independent of the query representation.
        return array_map(function ($specificPrice) {
            return (int) $specificPrice['from_quantity'];
        }, $specificPrices ?: []);
    }

    /**
     * Calculate the final price range for every country and group in one shop currency.
     *
     * @param int $idProduct
     * @param int $idShop
     * @param int $idCurrency
     * @param array $countries
     * @param array $groups
     * @param array $specificPriceQuantities
     * @param bool $useTax
     *
     * @return array
     */
    private function getPriceMatrix($idProduct, $idShop, $idCurrency, array $countries, array $groups, array $specificPriceQuantities, $useTax)
    {
        $prices = [];

        // Calculate each country and group combination through the core pricing logic so taxes and hooks affect dependency detection.
        foreach ($countries as $country) {
            $idCountry = (int) $country['id_country'];
            foreach ($groups as $group) {
                $idGroup = (int) $group['id_group'];
                $prices[$idCountry][$idGroup] = $this->calculatePriceRange($idProduct, $idShop, $idCurrency, $idCountry, $idGroup, $specificPriceQuantities, $useTax);
```

ps_facetedsearch.php
Codencode: @mattgoud 
Good catch on ANSI_QUOTES, thanks! I’ve switched those to single quotes.

The id_shop constraint is actually what this commit adds, so that part is already covered by ps.id_shop = psi.id_shop.

Codencode:
@mattgoud 
I’ve updated the join to use single quotes for `both` and `catalog`, so it’s safe with `ANSI_QUOTES`.

The `ps.id_shop = psi.id_shop` condition was already part of my commit and remains in place for the multistore cleanup.

PR #1320, opened 2026-09-05
Reject invalid values for the category filter depth

### LINKED ISSUE #36438: `ps_facetedsearch` : Invalid input are not considered as invalid
### Prerequisites

- [X] I understand and accept the project's [code of conduct](https://github.com/PrestaShop/PrestaShop/blob/develop/CODE_OF_CONDUCT.md).

For the input "Category filter depth", we add some inputs and they are not considered as invalid.

Input : 
* :red_circle:  : `-2`
* :red_circle:  : `2`
* :red_circle: :  : `LLOI`
* :red_circle: :  : `9223372036854775807`
* :red_circle: :  : `1,5`
* :red_circle: :  : `1 500`

Based on this scenario : 
* https://forge.prestashop.com/browse/TEST-4151
* https://build.prestashop-project.org/test-scenarios/scenarios/modules/ps-facetedsearch/configuration/faceted-search-module-category-filter-depth-field-configuration.html

### Expected behavior

_No response_

### Steps to reproduce

1. Go to "Modules > Module Manager" page
2. Search the module "ps_facetedsearch"
3. Go to the configuration page
4. Enter in the field an invalid value 
5. Save it 

9.0.x

8.1

ps_facetedsearch v3.16.1

@PrestaShopCorp

boo-code:
| Questions         | Answers
| Description?    | "Category filter depth" was a plain `<input type="text">` saved with a bare `(int)` cast, so nothing was ever rejected: `LLOI` became `0`, `1,5` became `1`, `1 500` became `1`, and `-2` was stored as a negative depth. The merchant ended up with a value they had not typed and no indication anything was wrong. This follows @jolelievre's direction on the issue - the input becomes `type="number"` with `min="0"` and `step="1"` so the browser validates inline, and the save path refuses anything that is not a whole number of zero or more instead of coercing it. Zero stays valid, since the field's own help text documents it as unlimited depth.
| Type?           | bug fix
| BC breaks?      | no
| Deprecations?   | no
| Fixed ticket?   | Fixes PrestaShop/PrestaShop#36438
| How to test?    | Module Manager > ps_facetedsearch > Configure. Type `-2`, `LLOI`, `1,5` or `1 500` in "Category filter depth" and save: the browser refuses the input, and a request that bypasses it is answered with "The category filter depth must be a whole number, zero or more." and nothing is stored. `0`, `1`, `2`, `3` still save normally.
| Sponsor company |

### Checked against every value in the report

```
'-2'                   rejected      '0'    accepted
'LLOI'                 rejected      '1'    accepted
'1,5'                  rejected      '2'    accepted
'1 500'                rejected      '3'    accepted
''                     rejected      ' 4 '  accepted (trimmed)
'1.0'  '007'  '1e3'    rejected
```

The module's PHPUnit suite passes with the change: **100 tests, 929 assertions, OK** (PHPUnit 5.7 on PHP 7.4, `tests/php/phpunit.xml`).

### One value the report lists that this does not reject

`9223372036854775807` is accepted, because it *is* a whole number of zero or more and the direction on the issue set no upper bound. Two things worth deciding rather than me inventing them: whether a maximum is wanted at all, and that without one the behaviour is platform dependent - on 64-bit PHP that string round-trips through `(int)` unchanged and passes, on 32-bit it does not and would be refused. If a bound is wanted, say which and I will add it.

### DIFF
--- a/ps_facetedsearch.php
+++ b/ps_facetedsearch.php
@@ -703,6 +703,18 @@ class Ps_Facetedsearch extends Module implements WidgetInterface
                 );
             }
         } elseif (Tools::isSubmit('submitLayeredSettings')) {
+            // The depth is a plain (int) cast below, which silently turns "LLOI" into 0, "1,5" into 1 and
+            // accepts negatives - so anything that is not a whole number of zero or more is rejected here
+            // rather than stored as something the merchant did not type.
+            $categoryDepth = Tools::getValue('ps_layered_filter_category_depth');
+            if (!is_numeric($categoryDepth) || (string) (int) $categoryDepth !== trim((string) $categoryDepth) || (int) $categoryDepth < 0) {
+                $this->context->smarty->assign('message', $this->displayError(
+                    $this->trans('The category filter depth must be a whole number, zero or more.', [], 'Modules.Facetedsearch.Admin')
+                ));
+
+                return;
+            }
+
             Configuration::updateValue('PS_LAYERED_CACHE_ENABLED', (int) Tools::getValue('ps_layered_cache_enabled'));
             Configuration::updateValue('PS_LAYERED_SHOW_QTIES', (int) Tools::getValue('ps_layered_show_qties'));
             Configuration::updateValue('PS_LAYERED_FULL_TREE', (int) Tools::getValue('ps_layered_full_tree'));
--- a/views/templates/admin/manage.tpl
+++ b/views/templates/admin/manage.tpl
@@ -202,7 +202,7 @@
 	<div class="form-group">
 	  <label class="col-lg-3 control-label">{l s='Category filter depth' d='Modules.Facetedsearch.Admin'}</label>
 	  <div class="col-lg-9">
-		<input type="text" name="ps_layered_filter_category_depth" value="{if $category_depth !== false}{$category_depth}{else}1{/if}" class="fixed-width-sm" />
+		<input type="number" min="0" step="1" name="ps_layered_filter_category_depth" value="{if $category_depth !== false}{$category_depth}{else}1{/if}" class="fixed-width-sm" />
 	  </div>
 		<div class="col-lg-9 col-lg-offset-3">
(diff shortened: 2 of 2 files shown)


boo-code:
The red `PHPStan (develop, 8.1|8.5)` here is not produced by this diff. Read from the two job logs:

- #1319, same check, green at `2026-09-04T08:34:47Z`: `Installing phpstan/phpstan (1.12.12)`, `extension-installer (1.2.0)`, no `phpstan-symfony`.
- this PR, red at `2026-09-05T09:59:05Z`: `Installing phpstan/phpstan (2.2.12)`, `extension-installer (1.4.3)`, `phpstan/phpstan-symfony (2.0.20)`.

The `develop` matrix installs PrestaShop's own dev dependencies, and PrestaShop/PrestaShop#42252 ("Update PHPStan from 1.x to version 2.2.6+") merged into `develop` at `2026-09-04T12:39:58Z` - between those two runs.

Ten errors are reported. Six name `UrlSegment` or `UrlSegmentValidator`: the validator's `validate()` signature, the `UnexpectedTypeException` thrown at `src/Constraint/UrlSegmentValidator.php:52`, and the four `new UrlSegment([...])` call sites in `src/Form/{Attribute,AttributeGroup,Feature,FeatureValue}/FormModifier.php`. `src/Constraint/UrlSegmentValidator.php` dates from 2019 and none of those files is touched here. The remaining four are two `is_array()` calls, a `str_replace()` and one more `is_array()`; this diff adds an `is_numeric()`, two casts and a comparison in `getContent()`, plus one input type in `manage.tpl`, so whichever lines those are, they are unchanged ones.

The module needs a pass over what PHPStan 2 with the Symfony extension now reports before anything can go green on that matrix. Widening `UrlSegmentValidator::validate()` to accept `mixed` and giving `UrlSegment::__construct()` a typed options parameter accounts for six of the ten.

PR #1324, opened 2026-09-08
Release 5.1.0

mattgoud:
| Questions         | Answers
| Description?      | Merge `dev` into `master` for release 5.1.0.
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | ~
| How to test?      | Install the module on a 8.2+ shop, then check the faceted search on the front office and the filter templates management in Back Office > Design > Faceted search: creating, editing and deleting a template must work, and prices must be indexed for a catalog with non-sequential product ids.
| Sponsor company   | @PrestaShopCorp

### Content

78 commits since `v5.0.0`, including four fixes touching shipped code:

- price indexation loop stopping early with non-sequential product ids
- edit and delete filter template links missing their token
- wrong language condition in a `LEFT JOIN` (`liaglv.id_lang` instead of `agl.id_lang`)
- position ordering guard for PrestaShop 8

The rest is dependency updates, coding style and CI.

### Status

The version bump to 5.1.0 landed on `dev` with PrestaShop/ps_facetedsearch#1325, so this merge now carries the right version number.

The two `PHPStan (develop, …)` jobs are red for a reason external to this release: PrestaShop `develop` moved from PHPStan 1.x to 2.x on 2026-09-04. PrestaShop/ps_facetedsearch#1327 addresses it. Every job analysing 8.2.x, 9.0.3, 9.1.5 and 9.2.x is green, and 9.2 is the target of this release.


mattgoud:
### Pre-QA, not a QA validation

Ran on my own PR, so this is exploratory pre-QA only: no `QA ✔️` label, and an independent QA pass is still required. Posting the findings so they are not lost.

Tested on: `dev` merged state, local Docker, PrestaShop **9.2.0**, single shop, module checked out on `bump-5.1.0`, 19 demo products.
No linked issue on this PR: it is a release merge, so the scope is the four fixes touching shipped code plus the module's Xray scenarios.

### Result per fix

| Fix | Verdict |
|---|---|
| Edit / delete / add filter template links missing their token | **OK** |
| Wrong language condition in the `LEFT JOIN` | **OK** |
| Position ordering guard for PrestaShop 8 | not exercised |
| Price indexation loop stopping early on non-sequential ids | **no verdict**, see below |

### TEST-8119 — show the number of matching products

Setting off, the 13 facet labels lose their counter; back on, all 13 show it again.



### TEST-8147 — use tax to filter price

The price facet moves from `19.00 - 36.00` to `22.00 - 44.00`. Worth knowing: the setting only takes effect after a **price index rebuild**, because `ps_layered_price_index` stores precomputed bounds. Flushing the block cache alone shows no change, which is misleading.



### TEST-8151 — edit template, and the token fix

The Edit link carries a `_token` and opens the form directly, no invalid-token interstitial. Name and category tree are correctly loaded.



### TEST-8179 — add new template

Same for the Add link: form opens with 13 filters available and nothing preselected.



### TEST-8161 / 8162 / 8164 / 8177 — filter types

All filter types are present in the template form with their value counts: Composition (6 values), Property (4 values), Color (14 attributes), Size (4 attributes), plus stock, extras, price, sub-categories, brand and condition.

filters

### TEST-8182 — BO header buttons

Back, Translate, Manage hooks and Help in the header; Index all missing prices, Rebuild entire price index, Build attributes and features indexes and Clear cache in the page. Cron tokens redacted.



### Price indexation: could not be reproduced

Built the fixture the bug needs: 149 products with ids up to 330, so ids well above the product count, and above the batch size of 100 so the loop actually iterates.

| State | Products indexed |
|---|---|
| With the fix | 149 / 149 |
| With the hunk reverted | 149 / 149 |

The reason is three lines below the loop: in non-ajax mode `indexPrices()` re-enters itself while the cursor advances, compensating for the early exit.

```php
if (($nbProducts > 0 && !$full || $cursor != $lastCursor && $full) && !$ajax) {
    return $this->indexPrices((int) $cursor, $full, $ajax, $smart);
}
```

So the CLI path cannot show the difference, and my attempt to drive the ajax path directly was malformed and returned a template render instead of JSON, which is a void measurement rather than a negative result. Settling it needs the BO "Rebuild entire price index" button in both states, or a unit test on the loop condition alone. Fixture removed afterwards, shop back to 19 products and all settings restored.

### Not observable on the demo catalog

- **TEST-8132** products from subcategories and **TEST-8140** category filter depth: the Clothes tree has a single level of children, so both settings produce the same facet
- **TEST-8138** out of stock last: every demo product is in stock

**Not run**: TEST-8075 upgrade, TEST-8134, TEST-8152.

### One trap worth knowing

The Module Manager **Update** action on a dev checkout downloads the published release and overwrites the working tree, while reporting `Update action on module ps_facetedsearch succeeded` and leaving the version in database unchanged. Use **Reset** when testing a checkout.

### Before merging

PrestaShop/ps_facetedsearch#1325 still has to land first, the version is `5.0.0` in both files on `dev`.

ps-jarvis:

    - [ ] `Checkbox (AND)`

cnavarro-prestashop (approved):

Quetzacoalt91 (approved):

cnavarro-prestashop:
# QA — `ps_facetedsearch` v4.0.3

**Environment:** PrestaShop 9.2.0, Hummingbird, single shop, EN, 19 products, 1 template (13 filters), Chrome.
**Source of truth:** the published [test-scenarios steps](https://build.prestashop-project.org/test-scenarios/scenarios/modules/ps-facetedsearch/configuration.html). 
**Out of scope:** module upgrade.

## Results

| Test | Scenario | Result |
| --- | --- | --- |
| [TEST-8119](https://prestashop-jira.atlassian.net/browse/TEST-8119) | Show the number of matching products | ✅ |
| [TEST-8151](https://prestashop-jira.atlassian.net/browse/TEST-8151) | Edit template — name and categories | ✅ |
| [TEST-8161](https://prestashop-jira.atlassian.net/browse/TEST-8161) | Edit template — Composition filter | ✅ |
| [TEST-8162](https://prestashop-jira.atlassian.net/browse/TEST-8162) | Edit template — Color filter | ✅ |
| [TEST-8164](https://prestashop-jira.atlassian.net/browse/TEST-8164) | Edit template — Property filter | ✅ |
| [TEST-8182](https://prestashop-jira.atlassian.net/browse/TEST-8182) | BO header buttons | ✅ |
| [TEST-8177](https://prestashop-jira.atlassian.net/browse/TEST-8177) | Edit template — Product weight filter | ❌ |
| [TEST-8132](https://prestashop-jira.atlassian.net/browse/TEST-8132) | Show products from subcategories | ⚪ not run |
| [TEST-8147](https://prestashop-jira.atlassian.net/browse/TEST-8147) | Use tax to filter price | ⚪ not run |
| [TEST-8152](https://prestashop-jira.atlassian.net/browse/TEST-8152) | Use rounding to filter price | ⚪ not run |

The three filter scenarios were run as one pass over six states (disabled → radio → drop-down → checkbox → limit 2 → no limit), verified on the front office after each save. All eighteen assertions held. Every save returned the expected `Your filter "[name]" was updated successfully.`

The three ⚪ open on preconditions that modify the shop outside the module — a product price plus a tax or rounding setting for 8147/8152, a new product for 8132. I stopped short: the new product could not be deleted afterwards, leaving a stray record in the catalogue.

## TEST-8177 — the failure

Disable → facet gone ✅. Re-enable → save accepted ✅ → **the weight facet never renders**, on any category.

Most likely all demo products carry weight 0, in which case suppressing the facet is correct and the *scenario* is at fault for lacking a precondition. I could not confirm it — the product edit page rejected direct URL access on an invalid CSRF token. Set distinct weights on two products to settle it.

## Defects found outside the scenarios

**1. Price settings do not propagate on save, and the fix is order-dependent.** Saving *Use tax to filter price* changes nothing in the front office. It takes **Rebuild entire price index, then Clear cache** — in that order. Clearing first re-caches from the stale index and leaves the front office wrong.

| Action | FO price range |
| --- | --- |
| baseline (tax = Yes) | €10.00 – €44.00 |
| set to No, Save | €10.00 – €44.00 |
| rebuild → clear | €9.00 – €36.00 |
| set to Yes, Save | €9.00 – €36.00 |
| clear → rebuild | €9.00 – €36.00 |
| rebuild → clear | €10.00 – €44.00 |

The template form and *Show the number of matching products* both propagate immediately, so this is specific to the price settings, not to the cache layer. The help text on *Enable cache system* claims invalidation on every modification, which is inaccurate here.

**2. Toggling a filter loses its front-office position.** Before: `… Size, Color, Composition, Property, Brand …`. After disabling and re-enabling those three: `… Size, Brand, Dimension, Paper Type, Color, Composition, Property`. They are appended at the end, and the template form offers no reordering, so this cannot be undone from the UI. It also means TEST-8161/8162/8164 permanently reorder the facets of any shop they run on.

**3. Wrong translation string.** Under *Categories used for this template*, the caption reads *"A customer account has already been created from this guest account. Please sign in."* — a front-office customer message used as a back-office field caption.

## Two things I got wrong on a first pass, for the record

The `New filters template` heading on the edit screen is **expected** — TEST-8151 asserts it twice. And the drop-down style is **not** broken: Hummingbird renders a Bootstrap dropdown, not a `<select>`, so the DOM looks empty until the menu is opened. Both were nearly filed as defects.

## Shop state

Configuration and template restored to their starting values, verified on a fresh reload. Front office back to 19 products, €10.00 – €44.00, counts displayed. **One residual change:** the facet order from defect 2, which cannot be reverted.

---

# Decision

**Do not sign off the campaign yet — but nothing here blocks the module.**

The module behaves correctly on everything that was testable: all six executed scenarios pass, including the full matrix of filter styles and result limits. No functional regression was found in the filtering itself.

What stands in the way of a sign-off is coverage and scenario quality rather than product defects:

- **TEST-8177 cannot pass as written** on a stock catalogue. Fix the scenario with a weight precondition, then re-run. Only if it still fails is there a bug.
- **Three scenarios remain unrun** and they cover the price and subcategory logic, which is where the one real defect surfaced. That is exactly the area least covered right now.
- **TEST-8147's own expected result** ends with *"(not now unfortunately)"*, an accepted deviation baked into the scenario. Resolve it before automating, or the automation will encode the ambiguity.

Recommended order: add the missing preconditions to 8177 and 8132, resolve the 8147 wording, run the three remaining scenarios, then open issues for defects 1 and 2. Defect 1 is the one worth prioritising — it is silent, it misleads the merchant, and the workaround is undiscoverable.

Defect 3 is cosmetic and can ride along with any translation pass.

PR #1327, opened 2026-09-08
Fix PHPStan 2 findings on the develop matrix

mattgoud:
| Questions         | Answers
| Description?      | Fix the PHPStan findings that appeared on the `develop` matrix after PrestaShop moved from PHPStan 1.x to 2.x, and align the module's composer PHP floor on the oldest PrestaShop it supports.
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | ~
| How to test?      | The two `PHPStan (develop, 8.1)` and `(develop, 8.5)` jobs must turn green, and every other PHPStan job plus PHPUnit must stay green.
| Sponsor company   | @PrestaShopCorp

### Why the CI turned red without the module changing

The workflow runs `../../vendor/bin/phpstan`, so the binary comes from the analysed PrestaShop branch. `develop` moved from PHPStan 1.x to 2.2 with PrestaShop/PrestaShop#42252, merged on 2026-09-04, and gained `phpstan/phpstan-symfony`.

Same module commit `1bc7c00`, two runs: green on 2026-08-29 with phpstan 1.12.12, red on 2026-09-08 with phpstan 2.2.12. The `9.2.x` matrix runs on the same PHP versions and stays green, since PrestaShop 9.2 still ships phpstan `^1.9.3`.

### Three real findings

- `ps_facetedsearch.php` lines 1216 and 1227, and `src/Product/CoreSearchBackport.php` line 393: `is_array()` on a value PHPStan 2 now infers as an array, so the condition is always true and is dropped.
- `src/Adapter/MySQL.php` line 536: `str_replace()` received an `int` inside an `array<string>`, now cast.

### Five findings that are an artefact, fixed by configuration

The remaining errors expected a type called `Symfony\Component\Validator\mixed`, which does not exist: it is the native `mixed` parsed as a class name, which only happens when the analysis assumes PHP below 8.0.

PHPStan takes the syntax level it parses with from the **platform** declaration, `config.platform.php` in `composer.json` plus `platform-overrides` in `composer.lock`. `require.php` is not read. The `develop` matrix only ever runs on PHP 8.1 and 8.5, so `phpVersion` is pinned in `tests/php/phpstan/phpstan-develop.neon` rather than rewriting the `UrlSegment` constraint and its four callers for a cause that is not in the module.

Raising the platform floor is **not** an alternative fix, and this was measured rather than assumed. On a branch carrying this PR without the `phpVersion` line and with the floor at `7.2.5`, the two `develop` jobs still fail with the same five findings while every other job stays green. `mixed` only became a native type in PHP 8.0, so no 7.x value removes them, and 8.0 is not reachable: `composer update --lock` refuses it because `phpunit/phpunit ~5.7` caps the dev dependencies below PHP 8.

### Composer PHP floor aligned on 7.2.5

Separately, `require.php` and `config.platform.php` both declared `>=7.1`, below the PHP floor of the oldest PrestaShop the module supports: `ps_versions_compliancy` min is `8.2.0` and PrestaShop 8.2 requires `>=7.2.5`. Nothing exercised 7.1 either, the syntax checker starts at 7.2 and PHPUnit runs on 7.4.

So `7.2.5` is simply the truthful value, it is the `PHP 7.2 to 8.5` target of PrestaShop/PrestaShop#41648, and it changes nothing for shops. The `composer.lock` diff is limited to `content-hash` and the two platform values.

### Note

`tests/php/phpstan/phpstan.neon` carries text accidentally pasted from the GitHub review UI on its first comment line. Harmless, not touched here to keep the diff focused.

### DIFF
--- a/tests/php/phpstan/phpstan-develop.neon
+++ b/tests/php/phpstan/phpstan-develop.neon
@@ -1,2 +1,9 @@
 includes:
 	- %currentWorkingDirectory%/tests/php/phpstan/phpstan.neon
+
+parameters:
+	# PHPStan 2 derives the analysed language level from composer.json ("php": ">=7.1").
+	# Under PHP 7.1 semantics, the native "mixed" type used by Symfony 7 is parsed as a
+	# class name, which reports Symfony\Component\Validator\mixed instead of mixed.
+	# The develop matrix only ever runs on PHP 8.1 and 8.5, so pin the level accordingly.
+	phpVersion: 80100
--- a/tests/php/phpstan/phpstan-develop.neon
+++ b/tests/php/phpstan/phpstan-develop.neon
@@ -2,8 +2,12 @@ includes:
 	- %currentWorkingDirectory%/tests/php/phpstan/phpstan.neon
 
 parameters:
-	# PHPStan 2 derives the analysed language level from composer.json ("php": ">=7.1").
-	# Under PHP 7.1 semantics, the native "mixed" type used by Symfony 7 is parsed as a
-	# class name, which reports Symfony\Component\Validator\mixed instead of mixed.
-	# The develop matrix only ever runs on PHP 8.1 and 8.5, so pin the level accordingly.
+	# PHPStan takes the PHP syntax level it parses with from the platform declaration,
+	# config.platform.php in composer.json (7.1.0 here) plus platform-overrides in
+	# composer.lock. Under 7.x semantics the native "mixed" type used by Symfony 7 is
+	# parsed as a class name, which reports Symfony\Component\Validator\mixed instead
+	# of mixed. "mixed" only became a type in PHP 8.0, so raising the platform floor
+	# within 7.x does not help, and 8.0 is not reachable while phpunit ~5.7 caps the
+	# dev dependencies below PHP 8. The develop matrix only ever runs on PHP 8.1 and
+	# 8.5, so pin the analysed level to what the job actually executes.
 	phpVersion: 80100
--- a/tests/php/phpstan/phpstan-develop.neon
+++ b/tests/php/phpstan/phpstan-develop.neon
@@ -3,11 +3,11 @@ includes:
 
 parameters:
 	# PHPStan takes the PHP syntax level it parses with from the platform declaration,
-	# config.platform.php in composer.json (7.1.0 here) plus platform-overrides in
+	# config.platform.php in composer.json (7.2.5 here) plus platform-overrides in
 	# composer.lock. Under 7.x semantics the native "mixed" type used by Symfony 7 is
 	# parsed as a class name, which reports Symfony\Component\Validator\mixed instead
-	# of mixed. "mixed" only became a type in PHP 8.0, so raising the platform floor
-	# within 7.x does not help, and 8.0 is not reachable while phpunit ~5.7 caps the
-	# dev dependencies below PHP 8. The develop matrix only ever runs on PHP 8.1 and
-	# 8.5, so pin the analysed level to what the job actually executes.
+	# of mixed. "mixed" only became a type in PHP 8.0, so no 7.x floor removes the
+	# findings, and 8.0 is not reachable while phpunit ~5.7 caps the dev dependencies
+	# below PHP 8. The develop matrix only ever runs on PHP 8.1 and 8.5, so pin the
+	# analysed level to what the job actually executes.
 	phpVersion: 80100
--- a/composer.json
+++ b/composer.json
@@ -10,7 +10,7 @@
         }
     ],
     "require": {
-        "php": ">=7.1",
+        "php": ">=7.2.5",
         "doctrine/collections": "^1.4"
     },
     "require-dev": {
@@ -20,7 +20,7 @@
     },
     "config": {
         "platform": {
-            "php": "7.1.0"
+            "php": "7.2.5"
         },
         "preferred-install": "dist",
         "prepend-autoloader": false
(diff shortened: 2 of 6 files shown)


jolelievre (dismissed):

Hlavtox (dismissed):

Quetzacoalt91 (dismissed):

cnavarro-prestashop (approved):

tests/php/phpstan/phpstan-develop.neon
```diff
@@ -1,2 +1,9 @@
 includes:
 	- %currentWorkingDirectory%/tests/php/phpstan/phpstan.neon
+
+parameters:
+	# PHPStan 2 derives the analysed language level from composer.json ("php": ">=7.1").
+	# Under PHP 7.1 semantics, the native "mixed" type used by Symfony 7 is parsed as a
+	# class name, which reports Symfony\Component\Validator\mixed instead of mixed.
+	# The develop matrix only ever runs on PHP 8.1 and 8.5, so pin the level accordingly.
+	phpVersion: 80100
```
cnavarro-prestashop, line 13: Instead changing PHP Version in phpstan config : change it in composer.json = 7.2 min instead 7.1

cnavarro-prestashop (changes requested):

tests/php/phpstan/phpstan-develop.neon
mattgoud: I measured that suggestion before answering, on a throwaway branch carrying this PR without the `phpVersion` line and with the platform floor moved to 7.2.5 (`config.platform.php` in `composer.json` and `platform-overrides` in `composer.lock`, via `composer update --lock`).

The two `develop` jobs still fail, with the same five findings:

```
Parameter #1 $value of class Symfony\Component\Validator\Exception\UnexpectedTypeException
  constructor expects Symfony\Component\Validator\Exception\mixed, ...Constraint given.
Parameter #1 $options of class PrestaShop\Module\FacetedSearch\Constraint\UrlSegment
  constructor expects Symfony\Component\Validator\mixed|null, array<string, string> given.   (x4)
```

Every other job stays green, so 7.2.5 changes nothing here. The reason is that `mixed` only became a native type in **PHP 8.0**: parsed under any 7.x semantics it is read as a class name, hence `Symfony\Component\Validator\mixed`. The floor would have to be 8.0, not 7.2.

And 8.0 is not reachable today. `composer update --lock` refuses it:

```
doctrine/instantiator 1.0.5 requires php >=5.3,<8.0-DEV
phpspec/prophecy v1.10.3 requires php ^5.3|^7.0
myclabs/deep-copy 1.7.0 requires php ^5.6 || ^7.0
```

The root is `phpunit/phpunit ~5.7` in `require-dev`, which caps the whole dev chain below PHP 8. Raising the platform means migrating PHPUnit first, which is a separate piece of work.

You were right on one point though, and I have pushed a fix for it: my comment in the neon said PHPStan derives the level from `require.php`, which it does not read. It is `config.platform.php` plus the lock's `platform-overrides` that drive the parsed syntax level. The comment now says so, and records why no 7.x floor helps.

Happy to open the PHPUnit upgrade as its own issue if you think the platform floor is worth raising on its own merits. For this PR the `phpVersion` pin states the truth: the `develop` matrix only ever runs on PHP 8.1 and 8.5.

mattgoud: Correction to my own last paragraph: I withdraw the offer to open a PHPUnit migration issue, it was the wrong suggestion and would have framed a much bigger decision as a tooling chore.

`config.platform.php` is a single knob that constrains the resolution of every dependency, dev and runtime alike. Raising it to 8.x to let a modern PHPUnit in would also let Composer pick **runtime** dependencies that a PHP 7.2 shop cannot execute. Since this module declares `ps_versions_compliancy` min `8.2.0`, and PrestaShop 8.2 itself requires PHP `>=7.2.5`, that floor has to stay in 7.x. An issue titled "migrate PHPUnit" would in practice mean "drop PrestaShop 8.2 support", which is another major and not a decision to smuggle in through a dev dependency.

Your suggestion is right for a different reason though, and I would rather apply it than argue it. The module currently declares `>=7.1` in both `require.php` and `config.platform.php`, which is **below the oldest PrestaShop it supports**. Nothing exercises that 7.1 either: the syntax checker starts at 7.2 and PHPUnit runs on 7.4. So 7.2.5 is simply the truthful value, and it is the Epic's stated "PHP 7.2 to 8.5" target, already applied to ps_cashondelivery and ps_wirepayment.

The two changes are independent and do not overlap, since an explicit `phpVersion` takes precedence over the platform declaration. I am measuring the combination on a throwaway branch right now and will push it here once the `develop` jobs confirm it:

| | value | why |
|---|---|---|
| `require.php` and `config.platform.php` | `7.1` -> `7.2.5` | truthful, PrestaShop 8.2 already requires it |
| `phpVersion` in the develop neon | `80100`, kept | that job only ever runs on PHP 8.1 and 8.5 |

mattgoud: Applied and green: 13 jobs out of 13, including `develop` on 8.1 and 8.5 plus PHPUnit.

Commit 360db51 moves `require.php` and `config.platform.php` from `>=7.1` to `>=7.2.5`, with `platform-overrides` regenerated in the lock. The lock diff is limited to `content-hash` and the two platform values, my local Composer had also rewritten `stability-flags`, `platform-dev` and downgraded `plugin-api-version`, which I reverted.

The `phpVersion` pin stays, for the reason measured above. The PR description now carries that measurement, and the earlier claim that PHPStan reads `require.php` is corrected there too.

Quetzacoalt91 (approved):

cnavarro-prestashop (approved):

